PR #62, opened 2023-09-06
Fixing 00 "bug"

matdenkas:
This turned out not to be a bug, but a complication of how we were doing things. Let me show you this graph.
Screenshot 2023-09-06 155917
I hope that worked! If not ill put it in the discord. Basically though, if you just take time range * interval, you create a block of time. The red rectangle. If its offset from the hour mark, then you get 12 hours of data points. However, if you back it up to the hour mark you get 13. This is a problem though as we would be expecting 12 data points and get 13 sometimes. Hence our bug.

The fix for this is basically just saying models have to be run at their iteration time. So if the model is hourly it HAS to be run on the hour (or the datepoint supplied must be on the hour). Basically we need to do it this way such that the DSPEC files and clearly communicate weather they want 12 points or 13 points. An hourly model wasn't designed to be run not on the hour as its closest data-point could only be on the hour its for. I get that this might be confusing and I'm happy to explain in person or on a call.

The code changes made will make it to where attempting to run a model without a properly aligned date time, will crash the system. Ill explain this more in the meeting.

### DIFF
--- a/data/dspec/test_dspec.json
+++ b/data/dspec/test_dspec.json
@@ -40,7 +40,7 @@
         "unit": "meter",
         "type": "float",
         "interval": 3600,
-        "range": [0, -12]
+        "range": [0, -11]
     },
     {
         "name": "x_wind_prev",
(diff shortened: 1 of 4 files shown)


bestrada33 (approved):
Looks good

lovelysandlonelys (approved):

PR #78, opened 2023-09-18
69 bug with lead time and time generated calculation

matdenkas:
So the issue was two things. First the test case had issues. An attribute was wrong and the date time was quering for the generated time not varification time. Secondly the way I was trying to calculate those in the query, did not work. So I fixed that for both quering outputs and predictions.


bestrada33:
Looks great and makes sense!

bestrada33 (approved):

lovelysandlonelys (approved):
the method of adjusting the time make sense to me, you look good to merge!

PR #87, opened 2023-09-27
Removing the check completely

matdenkas:
I removed the check completely there by letting requests not cause the error.


bestrada33:
Looks good

PR #100, opened 2023-10-19
97 update dspecs to work with new scema

bestrada33:
Test feels like overkill. Need to comment out the line below to run the test or else it gives an error. (Make sure to remove comment when done!)
image

Also, be sure to `pip install dotenv` if you haven't already. Isn't in this branches `requirements.txt`, but should be in the newer branches.


src/ModelExecution/inputGatherer.py
```diff
@@ -72,26 +72,39 @@ def __parse_dspec(self, dspecFileName: str) -> None:
...
                 input.location = inputJson["location"]
                 input.source = inputJson["source"]
                 input.series = inputJson["series"]
-                input.dataClassification = inputJson["dataClassification"]
-                input.unit = inputJson["unit"]
                 input.type = inputJson["type"]
-                input.datum = inputJson.get("datum")
                 input.interval = inputJson["interval"]
                 input.range = inputJson["range"]
+
```
matdenkas, line 97: The .get accessor returns none if the key is not found. Thus there is no reason to except a KeyError here

src/ModelExecution/dspec.py
```diff
@@ -46,23 +46,22 @@ def __str__(self) -> str:
...
-        self.dataClassification = None
         self.unit = None
         self.type = None
         self.datum = None
         self.interval = None
         self.range = None
 
     def __str__(self) -> str:
-        return f'[Input] -> name: {self.name}, location: {self.location}, source: {self.source}, series: {self.series}, dataClassification: {self.dataClassification}, unit: {self.unit}, type: {self.type}, datum: {self.datum}, range: {self.range}'
+        return f'[Input] -> name: {self.name}, location: {self.location}, source: {self.source}, series: {self.series}, type: {self.type}, datum: {self.datum}, range: {self.range}'
```
matdenkas, line 64: Only dataClassifications was removed from the object but both were removed from the __str__ method

src/ModelExecution/dspec.py
```diff
@@ -46,23 +46,22 @@ def __str__(self) -> str:
...
     
     def __repr__(self):
-        return f'[Input] -> name: {self.name}, location: {self.location}, source: {self.source}, series: {self.series}, dataClassification: {self.dataClassification}, unit: {self.unit}, type: {self.type}, datum: {self.datum}, range: {self.range}'
\ No newline at end of file
+        return f'[Input] -> name: {self.name}, location: {self.location}, source: {self.source}, series: {self.series}, type: {self.type}, datum: {self.datum}, range: {self.range}'
```
matdenkas, line 67: Same thing here, unit is still in the object

matdenkas (changes requested):
Up to you if you wanna fix these. They wont cause bugs.

src/ModelExecution/inputGatherer.py
bestrada33: Should I just use .get for unit too then? Instead of KeyError?

bestrada33: Nvm yeah using .get now instead

matdenkas: It is my understanding that yes you should

matdenkas (approved):
LGTM

abeasley1722 (approved):

PR #103, opened 2023-10-23
102 refactor inputgatherer to use new data objects and request format

bestrada33:
Test doesn't work yet but should work once SeriesProvider is refactored.


matdenkas (approved):

lovelysandlonelys (approved):
everything looks right to me

PR #105, opened 2023-10-23
104 refactor api endpoints

bestrada33:
Test doesn't work yet but should work once SeriesProvider is refactored.


matdenkas (approved):
That pytest thing with the api looks cool

lovelysandlonelys (approved):

PR #108, opened 2023-10-25
107 refactor seriesprovider to reflect new schema

matdenkas:
I had to change a lot in the request_input. We will need to talk about it at some point, because basically if you don't provide an interval it cant validate weather data is complete or not. This is highly important for dspecs as single point input specifications will have to still provide an interval! Which is weird!

I couldn't write testings thanks to relative imports. Ill bring it up at the meeting!


src/SeriesProvider/SeriesProvider.py
```diff
@@ -18,160 +18,119 @@
...
+                elif len(mergedResults) != amountOfExpectedResults:
+                    responseSeries.isComplete = False
+                    responseSeries.nonCompleteReason = 'Combined data ingestion and series storage results were more or less than expected.'
+                    return responseSeries
+                else: # This means the data is validated to be whole
+                    return responseSeries
+            
         except Exception as e:
-            return self.__get_and_log_err_response(requestDesc, [], f'An unknown error occurred attempting to fill request.\nRequest: {requestDesc}\nException: {format_exc()}')
+            errSeries = Series(seriesDescription, timeDescription, False)
```
bestrada33, line 99: Am I crazy or shouldn't `False` go before `timeDescription`? Assuming `False` is for `isComplete`

bestrada33 (changes requested):
https://github.com/conrad-blucher-institute/semaphore/pull/108/files#r1374021416

src/SeriesProvider/SeriesProvider.py
```diff
@@ -18,160 +18,119 @@
...
-            
-
-            ###Check contents contains the right amount of results
-
-            #Calculate the amnt of expected results
-            amntExpected = self.__get_amnt_of_results_expected(requestDesc)
-            if amntExpected is None:
-                return self.__get_and_log_err_response(requestDesc, dbSeries, f'Could not process series, {requestDesc.series}, interval value to determin amnt of expected results in request.')
+            # Attempt to pull request from series storage.
+            seriesStorageResults = self.seriesStorage.select_input(seriesDescription, timeDescription).data
```
bestrada33, line 60: Didn't know you could do .data at the end of a function call. How cool.

bestrada33 (approved):
Couldn't find anything obvious. Wish we could test. Should be fine though.

src/SeriesProvider/SeriesProvider.py
matdenkas: 🤓 Fluent Interface - https://en.wikipedia.org/wiki/Fluent_interface

PR #113, opened 2023-10-28
Move parse_date to utility.py. (Silly me.)

bestrada33:
Turns out that I was literally just remaking datetime.strptime. So just use datetime.strptime with the proper try-except block and should be able to handle any invalid inputs.


matdenkas (commented):
LGTM

abeasley1722 (approved):
No errors found

PR #123, opened 2023-11-03
79 create tcoon lighthouse ingestion class

matdenkas:
Make sure you have this mapping row to preform the tests

`insert_external_location_code('SouthBirdIsland', 'LIGHTHOUSE', '013', 0)`

### DIFF
--- a/src/DataIngestion/DataIngestion/LIGHTHOUSE.py
+++ b/src/DataIngestion/DataIngestion/LIGHTHOUSE.py
@@ -67,6 +67,11 @@ class LIGHTHOUSE(IDataIngestion):
         return data
 
     def __pull_pd_endpoint_dataPoint(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> Series | None:
+        """This function pulls data from LIGHTHOUSE's pd endpoint
+        :param seriesRequest: SeriesDescription - A data SeriesDescription object with the information to pull 
+        :param timeREquest: TimeDescription - A data TimeDescription object with the information to pull 
+        :param Series | None: A series containing the imported data or none if something went wrong
+"""
         
         # Reformat and sterilize datetimes
         fromString = timeDescription.fromDateTime.strftime('%m/%d/%y').replace('/', '%2F')
--- a/src/DataIngestion/DataIngestion/LIGHTHOUSE.py
+++ b/src/DataIngestion/DataIngestion/LIGHTHOUSE.py
@@ -23,34 +23,35 @@ SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
 sys.path.append(os.path.dirname(SCRIPT_DIR))
 
 from src.SeriesStorage.ISeriesStorage import series_storage_factory
-from DataClasses import Series, SeriesDescription, Actual, Prediction
+from DataClasses import Series, SeriesDescription, Input, TimeDescription
 from DataIngestion.IDataIngestion import IDataIngestion
 from utility import log
 
 from datetime import datetime
 from urllib.error import HTTPError
 from urllib.request import urlopen
-from math import radians, cos, sin
 import json
-
-from typing import List, Dict
 
 
 class LIGHTHOUSE(IDataIngestion):
+
     
     def __init__(self):
         self.seriesStorage = series_storage_factory()
+        self.lighthouseDataInfoMap = {'dAirTmp': ('atp', 'celsius'),
+                                       'dWaterTmp': ('wtp', 'celsius')}
 
     
-    def ingest_series(self, seriesDescription: SeriesDescription) -> Series | None:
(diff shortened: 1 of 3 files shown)


src/DataIngestion/DataIngestion/LIGHTHOUSE.py
```diff
@@ -0,0 +1,135 @@
+# -*- coding: utf-8 -*-
+#LIGHTHOUSE.py
+#----------------------------------
+# Created By: Matthew Kastl
+# Created Date: 4/8/2023
+# version 3.0
+#----------------------------------
+""" This file is a communicator with the NOAA tides and currents API. Each public method will provide the ingestion of one series from NOAA Tides and currents
```
bestrada33, line 8: Documentation needs to be updated!

src/tests/test_LIGHTHOUSE.py
```diff
@@ -0,0 +1,70 @@
...
+def test_ingest_series(seriesDescription: SeriesDescription, timeDescription: TimeDescription, expected_output: None):
+    """This function tests the test ingest series method. This test makes sure that the LIGHTHOUSE ingestion class is being properly mapped
+        function is complete.
+    """
+    load_dotenv()
+
+    lighthouse = data_ingestion_factory(seriesDescription)
+    result = lighthouse.ingest_series(seriesDescription, timeDescription)
+
+    assert result == expected_output
```
bestrada33, line 70: Wait so `expected_output` is `None` here?

src/DataIngestion/DataIngestion/LIGHTHOUSE.py
```diff
@@ -0,0 +1,135 @@
...
+from datetime import datetime
+from urllib.error import HTTPError
+from urllib.request import urlopen
+import json
+
+
+class LIGHTHOUSE(IDataIngestion):
+
+    
+    def __init__(self):
```
bestrada33, line 39: This doesn't need a `self.sourceCode` like the other ingestion classes?

src/tests/test_LIGHTHOUSE.py
```diff
@@ -0,0 +1,70 @@
...
+    (SeriesDescription('LIGHTHOUSE', 'apple', 'SouthBirdIsland'), TimeDescription(datetime.combine(date(2023, 9, 6), time(11, 0)), datetime.combine(date(2023, 9, 6), time(11, 0)), timedelta(seconds=3600)), None),
+])
+def test_pull_pd_endpoint_dataPoint(seriesDescription: SeriesDescription, timeDescription: TimeDescription, expected_output: int | None):
+    """This function tests the pull_pd_endpoint_dataPoint which is datapoints from LIGHTHOUSE's pd endpoint
+        function is complete.
+        NOTE:: This test depends on DBM.insert_external_location_code('SouthBirdIsland', 'LIGHTHOUSE', '013', 0) pre=existing in db
+    """
+    load_dotenv()
+
+    lighthouse = LIGHTHOUSE()
```
bestrada33, line 47: Why does this one not use `data_ingestion_factory`?

src/tests/test_LIGHTHOUSE.py
```diff
@@ -0,0 +1,70 @@
...
+    result = lighthouse._LIGHTHOUSE__pull_pd_endpoint_dataPoint(seriesDescription, timeDescription)
+
+    if result == None:
+        assert result == expected_output
+    else:
+        print(result.data)
```
bestrada33, line 53: Probably don't need this

src/DataIngestion/DataIngestion/LIGHTHOUSE.py
matdenkas: ✔️

src/DataIngestion/DataIngestion/LIGHTHOUSE.py
matdenkas: It can, its how you wanna set it up, now that things are so bundled nicely I decided to just keep it dynamic. There should be no possible way we reach this code with the sorce code being something else

src/tests/test_LIGHTHOUSE.py
matdenkas: Yes, this is only tests that the factory method is mapping ingestion requests to my DI class properly

src/tests/test_LIGHTHOUSE.py
matdenkas: ✔️ Good catch

src/tests/test_LIGHTHOUSE.py
matdenkas: Because this is only testing the ingestion method. If I use the facotry, im technically dragging too other methods in when I dont have to. If this test ever passes and the next test fails, that identifies the unit of the factory method as the issue and insures that the unit of this ingestion method as not being the issue

PR #135, opened 2023-11-09
Entrypoint

lovelysandlonelys:
A few things: 
- So, I wrote this with the assumption being that the database is already set up. I'm not sure if that should be the case or not? Since this is supposed to be like the easy way for new people to run semaphore there should probably be something that initializes the database right? But for this initial version run all the tests twice before testing arguments because we need the database.
-  I did attempt to write a test for this but I couldn't find much on mocking command line interfaces the way that I wrote mine so any thoughts on that would be appreciated!
- In the meeting last week there was a brief discussion about if the calculate reference time function should be placed in inputGatherer or somewhere else so maybe also keep that in mind? Joy and I put it where we thought it made most sense but we were thinking about a specific use case.

To Test: 
- python3 semaphoreRunner.py -h (for instructions)
- python3 semaphoreRunner.py -d test_dspec (to run test model now)


data/dspec/test_dspec.json
```diff
@@ -3,6 +3,11 @@
     "modelVersion": "1.0.0",
     "author": "John Doe",
     "modelFileName": "test_AI",
+    "timingInfo":{
```
bestrada33, line 6: I don't know what it is, but 'timing' throws me off. I can't think of a better word right now though.

semaphoreRunner.py
```diff
@@ -0,0 +1,89 @@
...
+        executionTime = datetime.now()
+
+        MW = ModelWrapper(fileName)
+        result = MW.make_and_save_prediction(executionTime)
+        print(result)
+        print(result.data)
+    else: 
+        MW = ModelWrapper(fileName)
+        result = MW.make_and_save_prediction(executionTime)
+        print(result)
```
bestrada33, line 38: print statements!!!

semaphoreRunner.py
```diff
@@ -0,0 +1,89 @@
...
+load_dotenv()
+
+def run_semaphore(fileName: str, executionTime: datetime = None):
+
+    if executionTime == None: 
+        executionTime = datetime.now()
+
+        MW = ModelWrapper(fileName)
+        result = MW.make_and_save_prediction(executionTime)
+        print(result)
```
bestrada33, line 33: Bunch of print statements! I think these should be logs?  Also, could probably just log `result` and `result.data` once after the if/else statements. Since no matter what they will be assigned

semaphoreRunner.py
```diff
@@ -0,0 +1,89 @@
...
+        args.dspec = args.dspec + '.json'
+    dspecFilePath = construct_true_path(getenv('DSPEC_FOLDER_PATH')) + (args.dspec )
+    
+    if not path.exists(dspecFilePath):
+        log(f'{dspecFilePath} not found!')
+        raise FileNotFoundError
+  
+    #if there is past time given
+    if args.past is not None: 
+        #checking that the past time passed is formatted correctly (stolen from beto)
```
bestrada33, line 71: Appreciate the shoutout! (capitalize the 'B' though ;)  )

semaphoreRunner.py
```diff
@@ -0,0 +1,89 @@
...
+        try: 
+            execution_time = datetime.strptime(args.past, '%Y%m%d%H%M')
+        except:
+            log('Bad datetime string given.')
+            raise ValueError('The provided time was not in the correct format.')
+        
+        #checking if date provided is in the past
+        if execution_time >= datetime.now():
```
bestrada33, line 79: Shouldn't this be `<` if it is checking if it is in the past? Instead of `>=`

semaphoreRunner.py
```diff
@@ -0,0 +1,89 @@
...
+        print(result)
+        print(result.data)
+    else: 
+        MW = ModelWrapper(fileName)
+        result = MW.make_and_save_prediction(executionTime)
+        print(result)
+        print(result.data) 
+
+#argument parsing
+def main(args=None):
```
bestrada33, line 42: Why is `args` optional? Is this like a thing? Genuinely asking because I don't know.

src/ModelExecution/inputGatherer.py
```diff
@@ -177,7 +187,18 @@ def __cast_data(self, data: list[Input], dataType: str) -> list[any]:
                     log(f'Input gatherer has no conversion for Unit: {dataType}')
                     raise NotImplementedError
         return castedData
+    
+    def calculate_referenceTime(self, execution :datetime) -> datetime:
```
bestrada33, line 191: This seems to work well

data/dspec/test_dspec.json
lovelysandlonelys: vibe. we kinda just named it the first thing we thought of but i can see how it might be kinda vague. maybe something like intervalSpecifications? since it's mostly just information on interval?

semaphoreRunner.py
lovelysandlonelys: o7

data/dspec/test_dspec.json
bestrada33: Ahhhhh, maybe it won't matter. All the rest in dspec.py are {blahblah}Info anyways.

bestrada33 (changes requested):
This branch should be fine, just the small changes I commented on.

I made an item for the `post_process_prediction` bugs in the Projects Board

semaphoreRunner.py
lovelysandlonelys: so, i made this change when i was looking for a way to write a test for the runner and all the examples I found did this so that they could call the function in the testing file

src/semaphoreRunner.py
```diff
@@ -0,0 +1,89 @@
...
+
+from src.ModelExecution.modelWrapper import ModelWrapper
+from src.utility import log, construct_true_path
+from datetime import datetime
+
+load_dotenv()
+
+def run_semaphore(fileName: str, executionTime: datetime = None):
+
+    if executionTime == None:
```
abeasley1722, line 28: I've read that we should use "is" when comparing python constants like None

matdenkas: It is a better practice, Granted I do == None all the time but when have I ever done anything the right way. Its pure optimization though, saving a call to an __eq__. All that means is that doing == None and is None will give you the same result in every situation, just is is faster as it will check typing information rather than data.

matdenkas: I hope that toneyly comes across right

matdenkas (changes requested):
`PS C:\Dev\semaphore> python3 ./src/semaphoreRunner.py
Traceback (most recent call last):
  File "C:\Dev\semaphore\src\semaphoreRunner.py", line 20, in <module>
    from src.ModelExecution.modelWrapper import ModelWrapper
ModuleNotFoundError: No module named 'src'
PS C:\Dev\semaphore>`

You need to adjust import statements when you move files

matdenkas (approved):

PR #145, opened 2023-11-13
Updated .dockerignore

bestrada33:
Based it mostly off of the .gitignore file and what I saw online.

- docs and db were already on there. What are they for?
- Let me know what doesn't need to be ignored and what is missing.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,42 @@
-.env
 **/*.pyc
 docs
 db
+
+# VSCode
 .vscode
-data
+
+# .env files
+*.env
+
+# Ignore virtual enviroment root
+venv/
+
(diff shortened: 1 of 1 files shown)


snguyen1:
Do we need files in the data directory there when we package into a docker image? Or is it going to be mounted to a volume on host machine and read the files from there dynamically?

bestrada33:
> Do we need files in the data directory there when we package into a docker image? Or is it going to be mounted to a volume on host machine and read the files from there dynamically?

Dynamically may be better. So we don't have to push a new image every time we get a new DSPEC or h5 file.

bestrada33:
Now ignoring data directory

PR #146, opened 2023-11-17
127 operationalize thermal refuge model

matdenkas:
Note a change here where we tell tenserflow to not compile the model. This is because compiling a model can include custom functions, like custom loss functions for example. We don't have these but we dont need them. We just want to run the model not the machine that trained it. So for every model this flag should be set to false! 

Watch out for custom functions and namespaces. I really really really hope this will never ever come up again, but if there is a way to have a custom function before or after the model then this is going to be something we will need to support.


src/DataIngestion/DataIngestion/LIGHTHOUSE.py
```diff
@@ -125,5 +129,5 @@ def __pull_pd_endpoint_dataPoint(self, seriesDescription: SeriesDescription, tim
         resultSeries.data = inputs
         
         #Insert series into DB
-        self.seriesStorage.insert_input(resultSeries)
+        self.seriesStorage.insert_input(resultSeries).data
```
bestrada33, line 132: Should this be done for every ingestion class?

src/DataIngestion/DataIngestion/LIGHTHOUSE.py
```diff
@@ -108,7 +108,11 @@ def __pull_pd_endpoint_dataPoint(self, seriesDescription: SeriesDescription, tim
                 epochTimeStamp = dataPoint[dataTimestampIndex]/1000
                 if timeDescription.interval != None:
                     if epochTimeStamp % timeDescription.interval.total_seconds() != 0:
-                        continue
+                        continue    
+
+                # Lighthouse over returns data, so we just clip any data that is before or after our requested date range.
```
bestrada33, line 113: Will probably need to do this for NDFD.

bestrada33 (approved):
I mean it looks good. Is there a way to test it?

src/DataIngestion/DataIngestion/LIGHTHOUSE.py
matdenkas: Yes :3

PR #164, opened 2023-12-18
Adding method to check db exists and integrating it with the init db …

matdenkas:
…script


lovelysandlonelys (approved):
this makes sense to me!

abeasley1722:
LGTM

abeasley1722 (approved):

tools/init_db.py
```diff
@@ -31,16 +30,23 @@ def readInitCSV(csvFileName: str) -> list:
...
-    # Insert reference and mapping data
-    sqlorm.insert_ref_dataDatum(readInitCSV('dataDatum.csv'))
-    sqlorm.insert_ref_dataLocation(readInitCSV('dataLocation.csv'))
-    sqlorm.insert_ref_dataSeries(readInitCSV('dataSeries.csv'))
-    sqlorm.insert_ref_dataSource(readInitCSV('dataSource.csv'))
-    sqlorm.insert_ref_dataUnit(readInitCSV('dataUnit.csv'))
-    sqlorm.insert_data_mapping(readInitCSV('dataMapping.csv'))
+
+
+    dbEists = sqlorm.DB_exists()
```
bestrada33, line 35: Guessing this should be dbExists? Missing the 'x'

tools/init_db.py
```diff
@@ -31,16 +30,23 @@ def readInitCSV(csvFileName: str) -> list:
...
+    if dbEists == -1: # Clear out all data if some tables but not all tables are found
+        sqlorm.drop_DB()
+        dbEists = 0
+
+    if dbEists == 0: # Make the DB
+        sqlorm.create_DB()
+
+        # Insert reference and mapping data
+        sqlorm.insert_ref_dataDatum(readInitCSV('dataDatum.csv'))
```
bestrada33, line 44: So, if the DB exists and has all tables it just overwrites the reference data no matter what? Even if the reference data was already in there?

bestrada33 (approved):
LGTM. Seems to work well. Only left one comment, but should be fine

PR #169, opened 2024-01-01
168 entry point has argument to not save the computed value

matdenkas:
image
Running without the toss flag stores the prediction in the DB

image
Running with the toss flag prevents storage in the DB


bestrada33 (approved):
LGTM!

lovelysandlonelys (commented):
works on my machine as well! looks good

PR #181, opened 2024-01-12
added timing info where needed and removed dataClasification

lovelysandlonelys:
I think that really there was only the one remnant of the old dspec in MLP with the dataClassification, everything else was included in the despc.py file. I wasn't completely sure what timing info to put for the thermal refuge so I just put hourly.


matdenkas (commented):
LGTM

PR #196, opened 2024-02-05
Fixing crash on LIGHTHOUSE returning all null values

matdenkas:
LIGHTHOUSE was crashing when the PD returned only nulls. When we tried to insert the empty list into the DB the crash happens. I fixed this by having the LIGHTHOUSE class check and see if no good inputs were found, and if so, it does not insert into the DB and logs what happened.


src/DataIngestion/DataIngestion/LIGHTHOUSE.py
```diff
@@ -86,6 +86,7 @@ def __pull_pd_endpoint_dataPoint(self, seriesDescription: SeriesDescription, tim
         ### Get LIGHTHOUSE data
         # Build the URL to hit
         url = f'http://lighthouse.tamucc.edu/pd?stnlist={lighthouseLocationCode}&serlist={sim[SIMSeriesCodeIndex]}&when={fromString}%2C{toString}&whentz=UTC0&-action=app_json&unit=metric&elev=stnd'
+        print(url)
```
bestrada33, line 89: Print statement!!!!

bestrada33 (changes requested):
Print statement spotted. Left a comment at its location

abeasley1722 (approved):
LGTM

PR #198, opened 2024-02-05
191 update imports

matdenkas:
I redid the imports (except for tests) in a more proper and easier manner. I also refactored some file and folder names. Finally I had to update some docker stuff to get the API to play nice with the new imports. This should save many headaches moving forwards


src/ModelExecution/OH_Classes/OnePackedFloat.py
```diff
@@ -13,13 +13,9 @@
...
 #
 #Imports
-import sys
-import os
-SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__)) 
-sys.path.append(os.path.dirname(SCRIPT_DIR))
 from DataClasses import Output
 from ModelExecution.IOutputHandler import IOutputHandler
-from dspec import Dspec
+from ..dspec import Dspec
```
bestrada33, line 18: This works? Does it run outside of the containers? Is this best practice?

src/ModelExecution/__init__.py
bestrada33, line 1: What exactly does this do?

abeasley1722 (changes requested):
The tools need to be updated. Getting the "DataClass" module doesn't exist error when I run init_db.py. I believe everything else is good!

PR #208, opened 2024-02-16
193 update series provider merge method

ccftissot:
Creating this request so I can comment directly in the code

### DIFF
--- a/src/SeriesProvider/SeriesProvider.py
+++ b/src/SeriesProvider/SeriesProvider.py
@@ -169,9 +169,8 @@ class SeriesProvider():
         # GenerateTimeStamp will calculate a timestamp an amount of steps away from the initial time
         generateTimestamp = lambda initial_time, idx, interval : initial_time + (interval * idx)
 
-        # Preform list comprehension to generate a list of all the time steps we need
+        # Preform list comprehension to generate a list of all the time steps we need plus another list of the same size this is all None
         keys = [generateTimestamp(initial_time, idx, timeDescription.interval) for idx in range(steps)]
-        # Generate a list of nones the same length
         values = [None] * len(keys)
 
         # zip both lists in a dictionary and return the result
--- a/src/SeriesProvider/SeriesProvider.py
+++ b/src/SeriesProvider/SeriesProvider.py
@@ -44,7 +44,7 @@ class SeriesProvider():
         """This method attempts to return a series matching a series description and a time description.
             It will attempt first to get the series from series storage, kicking off data ingestion if series storage
             doesn't have all the data.
-            NOTE: If an interval is not provided in the time description, the interval will be assumed to be 6 minuets
+            NOTE: If an interval is not provided in the time description, the interval will be assumed to be 6 minutes
             :param seriesDescription - A description of the wanted series.
             :param timeDescription - A description of the temporal information of the wanted series. 
             :returns series - The series containing as much data as could be found.
@@ -60,7 +60,7 @@ class SeriesProvider():
             return validated_series_storage_results
 
         # If series storage results weren't valid
-        # Pull data ingestion validate it with the series storage results, if valid return it
+        # Pull data ingestion, validate it with the series storage results, if valid return it
         data_ingestion_class = data_ingestion_factory(seriesDescription)
         data_ingestion_results = data_ingestion_class.ingest_series(seriesDescription, timeDescription)
         validated_merged_result = self.__generate_resulting_series(seriesDescription, timeDescription, series_storage_results.data, data_ingestion_results.data)
@@ -113,7 +113,7 @@ class SeriesProvider():
         # Construct a dictionary for the db results
         database_results = { input.timeVerified : input for input in DBList}
 
-        # If there are data ingestion results construct a dictionary for that oo
+        # If there are data ingestion results construct a dictionary for that too
         if DIList != None:
             ingestion_results = { input.timeVerified : input for input in DIList}
 
@@ -138,7 +138,7 @@ class SeriesProvider():
             isComplete = False
             reason_string = f'There were {missing_results} missing results!'
 
-            # If there were missing values, that cause's some None's to remain. We clear them out in this step
+            # If a result was missing, an input was not mapped to that key. Thus its mapped to None. We remove the whole k:v pair as to not pollute the results with None.
             datetimeDict_filtered = {k: v for k, v in datetimeDict.items() if v is not None}
             datetimeDict.clear()
             datetimeDict.update(datetimeDict_filtered)
--- a/src/SeriesProvider/SeriesProvider.py
+++ b/src/SeriesProvider/SeriesProvider.py
@@ -13,7 +13,7 @@
 #Imports
 from SeriesStorage.ISeriesStorage import series_storage_factory
 from DataIngestion.IDataIngestion import data_ingestion_factory
-from DataClasses import Series, SemaphoreSeriesDescription, SeriesDescription, TimeDescription
+from DataClasses import Series, SemaphoreSeriesDescription, SeriesDescription, TimeDescription, Input
 from utility import log
 
 from datetime import timedelta
@@ -55,7 +55,7 @@ class SeriesProvider():
         
         # Pull data from series storage and validate it, if valid return it
         series_storage_results = self.seriesStorage.select_input(seriesDescription, timeDescription)
-        validated_series_storage_results = self.__generate_resulting_series(seriesDescription, timeDescription, series_storage_results)
+        validated_series_storage_results = self.__generate_resulting_series(seriesDescription, timeDescription, series_storage_results.data)
         if(validated_series_storage_results.isComplete):
             return validated_series_storage_results
 
@@ -63,7 +63,7 @@ class SeriesProvider():
         # Pull data ingestion validate it with the series storage results, if valid return it
         data_ingestion_class = data_ingestion_factory(seriesDescription)
         data_ingestion_results = data_ingestion_class.ingest_series(seriesDescription, timeDescription)
-        validated_merged_result = self.__generate_resulting_series(seriesDescription, timeDescription, series_storage_results, data_ingestion_results)
+        validated_merged_result = self.__generate_resulting_series(seriesDescription, timeDescription, series_storage_results.data, data_ingestion_results.data)
         if(validated_merged_result.isComplete):
             return validated_merged_result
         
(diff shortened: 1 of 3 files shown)


src/SeriesProvider/SeriesProvider.py
```diff
@@ -105,38 +90,89 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
+            description= seriesDescription, 
+            isComplete= isComplete,
+            timeDescription=timeDescription,
+            nonCompleteReason=reason_string
+        )
+        result.data = list(datetimeDict.values())
+        return result
+                
+ 
+    def __generate_datetime_map(self, timeDescription: TimeDescription) -> dict:
```
ccftissot, line 153: consider making this method more straight forward.  In general it is good practice not to assume what the caller of a function will do with the results.  Making it return a dictionary with null value is a awkward.  Consider making it return a list of timestamp instead.  Our calling function only uses the dictionary to iterate through its keys anyway.  Rename the method  __get_datetime_list or something like that.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -105,38 +90,89 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
 
-        #TODO:: This is a very slow, it was optimized with hashing like this:    
-            # uniqueToSecond = set(second) - set(first)
-            # return first + list(uniqueToSecond)
-        #But its not hashing properly anymore, needs to be looked into
+        # Iterate over every timestamp we are expecting 
+        for datetime in datetimeDict.keys():
+            
+            # Search the DB results for every input that matches that timestamp
+            valid_database_results = [input for input in DBSeries.data if input.timeGenerated == datetime]
```
ccftissot, line 113: For improved performance, consider generating dictionaries for both lists outside of the loop and catching duplicates in each list at that time.  The dictionaries will be keyed by timestamps.  Generating a dictionary from a list will cause a single full iteration on the list.  The way the code is written right now, we are doing one full iteration on the first list for every timestamp we are interested in.  One you generated the dictionary, you can quickly and easily check whether an entry for a given timestamp exist in the list.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -105,38 +90,89 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
+            if len(valid_database_results) > 1:
+                log(f'Warning: DB results returned multiple results for the same datetime \n{DBSeries}\n{valid_database_results}')
```
ccftissot, line 115: are there scenarios where this could happen?  I think we need to make sure that this should not happen, and if it does, this should constitute an error and we should throw an error rather than just logging.  When we ask the persist layer to retrieve our inputs, it should do so by always getting the ones with the most recent time acquired timestamp.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -105,38 +90,89 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
-        """
-
-        if timeDescription.interval == None: return None
 
-        totalSecondsRequested = (timeDescription.toDateTime - timeDescription.fromDateTime).total_seconds()
 
-        # We add one to acount for this being an inclusive time range. A time range from 12pm - 2pm includes both 12pm and 2pm.
-        return 1 + totalSecondsRequested // timeDescription.interval.total_seconds()
-    
+    def __validate_results(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription, DBSeries: Series, DISeries: Series = None) -> Series:
```
ccftissot, line 95: consider renaming this method as it does more than just validate results.  maybe something like generate_result_series or something similar.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -105,38 +90,89 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
+                log(f'Warning: DB results returned multiple results for the same datetime \n{DBSeries}\n{valid_database_results}')
+
+            # Search the DI results for every input that matches that timestamp if there are DI results
+            if DISeries != None:
+                valid_ingestion_results = [input for input in DISeries.data if input.timeGenerated == datetime]
+                if len(valid_ingestion_results) > 1:
+                    log(f'Warning: DataIngestion results returned multiple results for the same datetime \n{DISeries}\n{valid_ingestion_results}')
+
+            # If there are no results from either DB or DI thats missing, 
+            # if there are results we prioritize the DI result as thats assumed to be more accurate
```
ccftissot, line 124: consider using time_acquired instead and use whichever is the most recently acquired value

src/SeriesProvider/SeriesProvider.py
```diff
@@ -105,38 +90,89 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
+            # Search the DB results for every input that matches that timestamp
+            valid_database_results = [input for input in DBSeries.data if input.timeGenerated == datetime]
+            if len(valid_database_results) > 1:
+                log(f'Warning: DB results returned multiple results for the same datetime \n{DBSeries}\n{valid_database_results}')
+
+            # Search the DI results for every input that matches that timestamp if there are DI results
+            if DISeries != None:
+                valid_ingestion_results = [input for input in DISeries.data if input.timeGenerated == datetime]
+                if len(valid_ingestion_results) > 1:
+                    log(f'Warning: DataIngestion results returned multiple results for the same datetime \n{DISeries}\n{valid_ingestion_results}')
```
ccftissot, line 121: is this a valid scenario?  if not, we should throw an error

src/SeriesProvider/SeriesProvider.py
```diff
@@ -105,38 +90,89 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
+
+            # If there are no results from either DB or DI thats missing, 
+            # if there are results we prioritize the DI result as thats assumed to be more accurate
+            DB_result_exists = len(valid_database_results) > 0
+            DI_result_exits = DISeries != None and len(valid_ingestion_results) > 0
+
+            if not DB_result_exists and not DI_result_exits:
+                missing_results = missing_results + 1
+            elif DI_result_exits:
+                datetimeDict[datetime] = valid_ingestion_results[0]
```
ccftissot, line 131: seems a little arbitrary to just pick the first one, no?

src/tests/IntegrationTests/test_SeriesProvider.py
```diff
@@ -65,32 +65,12 @@ def test_request_output():
...
-    (datetime(2001, 12, 28), datetime(2001, 12, 28) + timedelta(hours=24), timedelta(hours=1), 25), # interval < fromTime - toTime
-    (datetime(2001, 12, 28), datetime(2001, 12, 28) + timedelta(hours=24), timedelta(hours=24), 2), # Time range = fromTime - toTime
-    (datetime(2001, 12, 28), datetime(2001, 12, 28), timedelta(hours=24), 1) # single point interval > fromTime - toTime
-    ])
-def test__get_amount_of_results_expected(fromTime: datetime, toTime: datetime, interval: timedelta, expected_output: int):
-    """Tests the get amount of results expected that it is correctly calculating correct amount
-    of expected results given a to time and from time and some interval
-    """
-    load_dotenv()
+@pytest.mark.parametrize("timeDescription, expected_output", [
```
ccftissot, line 68: consider adding a couple of realistic scenarios for example for inputs with an interval of 6 minutes.  Maybe use examples from the cold stunning model and the thermal refuge model

src/tests/IntegrationTests/test_SeriesProvider.py
ccftissot, line 1: Consider adding a test method to test the validate_results method.  I would recommend covering the following scenarios:
1. a single list is passed in and is complete (has all required timestamps) with no duplicate
2. a single list is passed in and is complete with some duplicates
3. a single list is passed in and is incomplete (missing timestamps) with no duplicate
4. two lists are passed in. the first one is incomplete but all missing timestamps can be found in the second list.  The second list does not have any duplicates.
5. two lists are passed in. the first one is incomplete but all missing timestamps can be found in the second list.  The second list does have internal duplicates but is not duplicating entries from the first list (or the values are the same)
6. two lists are passed in. the first one is incomplete and the second list has duplicate timestamp entries of the first list with different values.  check that the correct values are picked by the method.
7.  etc.

ccftissot (commented):
hey @matdenkas. i am done with the review.  let me know if you want to discuss.

src/tests/IntegrationTests/test_SeriesProvider.py
matdenkas: All 6+ cases handled

src/SeriesProvider/SeriesProvider.py
matdenkas: Solved by switching to dicts

src/SeriesProvider/SeriesProvider.py
matdenkas: DI values are always freashly gathered, thus those will always be the most recently acquired value, updated comment

src/SeriesProvider/SeriesProvider.py
matdenkas: Solved by switching to dicts

src/SeriesProvider/SeriesProvider.py
matdenkas: Done

src/SeriesProvider/SeriesProvider.py
```diff
@@ -41,58 +44,40 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
         """This method attempts to return a series matching a series description and a time description.
             It will attempt first to get the series from series storage, kicking off data ingestion if series storage
             doesn't have all the data.
-            NOTE: If an interval is not provided in the time description, the isComplete flag will always be false unless its a single data point (toTime = fromTime)
+            NOTE: If an interval is not provided in the time description, the interval will be assumed to be 6 minuets
```
abeasley1722, line 47: minutes*

src/SeriesProvider/SeriesProvider.py
```diff
@@ -13,7 +13,10 @@
 #Imports
 from SeriesStorage.ISeriesStorage import series_storage_factory
 from DataIngestion.IDataIngestion import data_ingestion_factory
```
abeasley1722, line 15: Tests ran as they should. Looks good to me.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -105,38 +90,89 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
-            # return first + list(uniqueToSecond)
-        #But its not hashing properly anymore, needs to be looked into
+        # Construct a dictionary for the db results
+        database_results = { input.timeVerified : input for input in DBList}
 
-        for item in second:
-            if item not in first:
-                first.append(item)
-        return first
+        # If there are data ingestion results construct a dictionary for that oo
```
lovelysandlonelys, line 116: lol "construct a dictionary for that oo" probably mean too here

abeasley1722 (approved):

src/SeriesProvider/SeriesProvider.py
```diff
@@ -105,38 +90,89 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
+
+        # If there were missing results we assign as such in the series
+        # No log here, as this method will be used to also detect if Data Ingestion should be kicked off
+        isComplete = True
+        reason_string = ''
+        if missing_results > 0:
+            isComplete = False
+            reason_string = f'There were {missing_results} missing results!'
+
+            # If there were missing values, that cause's some None's to remain. We clear them out in this step
```
lovelysandlonelys, line 141: #If there were any missing values in the merged results, there are some None values in the dictionary. We clear them out in this step

if i understand this right this might be a better way of phrasing this ^

lovelysandlonelys (approved):

PR #211, opened 2024-02-21
209 add better logging

matdenkas:
The DB errors weren't being reported not because they were caught but because they weren't caught. I added a proper try except case at the very top of running semaphore, in semaphoRerunner.py. That way we will be logging these errors.

I also added logging at major points

Blocked by merge method PR: 193


lovelysandlonelys (approved):
good logging!

PR #216, opened 2024-02-22
215 series provider remake not handling nones correctly

matdenkas:
Fixed by checking if its none before we try and access data


lovelysandlonelys (approved):
makes sense to me 👍🏻

abeasley1722 (approved):
Looks good to me!

PR #262, opened 2024-04-03
223 data base account follow up

lovelysandlonelys:
Hello! 

I have added code to the init_db.py file that adds an API user and changed the location string in the yml file that the api container uses to use this account information. I also made a script that we can run to create admin/researcher accounts, I think that the plan for user accounts was to add manually so that's what that's for, just to make it easier. 

Let me know if anything looks strange and please try to test, I used the api with the new permissions to pull inputs and it did so successfully with the permissions i assigned. To check what the permissions are in PG Admin you can use the PSQL tool and the command \z to check the permissions of each user on all of the tables.


abeasley1722 (approved):
Worked as expected!

lovelysandlonelys:
Tissot: Bad Practice: We don't want to concatenate strings with input variables, it leaves the function open to SQL injection attacks.

lovelysandlonelys:
We also need to see how we can keep the api account across deployments, since we don't want to run init_db again once the database has been up for a while

PR #271, opened 2024-04-11
Fixing NAN 4mm bug

matdenkas:
Fixed by filtering any nones from lists before preforming the sum operation.
The sum operation returns Nan if non numeric data is within the list.
We clear Nones before selecting the four max to make sure we select the 4 highest values that we actually have.

### DIFF
--- a/src/DataIngestion/DI_Classes/NOAATANDC.py
+++ b/src/DataIngestion/DI_Classes/NOAATANDC.py
@@ -248,7 +248,8 @@ class NOAATANDC(IDataIngestion):
         if data is None: return None
 
         data = data['v'].values
-        four_highest = sorted(data)[-4:]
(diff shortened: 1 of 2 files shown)


src/DataIngestion/DI_Classes/NOAATANDC.py
```diff
@@ -248,7 +248,8 @@ def __fetch_4_max_mean_dWl(self, seriesDescription: SeriesDescription, timeDescr
         if data is None: return None
 
         data = data['v'].values
-        four_highest = sorted(data)[-4:]
+        input_data = list(filter(lambda item: item is not None, data)) # Removing any Nones from the list
```
abeasley1722, line 251: Makes sense to me.

abeasley1722 (approved):
LGTM

bestrada33 (approved):
LGTM

PR #280, opened 2024-04-14
249 fix log file

abeasley1722:
1. Creation of a Logs Directory: A new directory named logs will be created at the root path.

2. Logs Organized by Year and Month: Within the log directory, logs for each model will be organized based on the year and month they were generated.

3. Additional log file added for Cron Logs: I added an additional log file specifically for cron jobs associated with each model. Crontabs produce two separate outputs: one for command execution and another highlighting general errors encountered during the cron job execution. To improve error tracking and management, general crontab error messages will be sent to a cron log file instead of heirloom mail. I believe it will be easier to navigate in the future rather than the heirloom mail system it was being sent to previously.


lovelysandlonelys (approved):
Looks good to me, good job figuring out the escape character for the percentages 👍🏻

PR #299, opened 2024-04-28
284 Operationalization of all Inundation months and small change of modelwrapper to reflect tf layer changes

abeasley1722:
### Changes:

1. Created the dspec for each month of the inundation models
2. Shortened the names of the dspecs for each model as some exceeded the character limit that can be added to the database
3. Added each model to the crontab
4. Changed a small part of the modelwrapper as we can no longer get input_shapes from layers in the newer versions of tensorflow. firstlayer and last layer were not used in any other parts of the code, so I deleted them. This should also fix the recent "Dense" bug. 

If crontab does not work on your WSL just run the "docker exec command... " from the crontab in the parent directory of Semaphore.


data/dspec/Inundation/August/ar_inundation_aug_12h.json
```diff
@@ -0,0 +1,63 @@
+{
+    "modelName": "ar_inundation_aug_12h",
+    "modelVersion": "1.0.1",
```
lovelysandlonelys, line 3: Isn't this technically version 1.0.0 because we've never had the inundation models loaded before?

src/ModelExecution/modelWrapper.py
```diff
@@ -67,11 +68,12 @@ def __shape_data(self, inputs: array) -> None:
         to the correct shape required by the model, dynamically. The shape that the input comes in
         and what the model was trained with, aren't necessarily the same.
         """
-
-        #Get only first and last layers
-        firstLayer, *_, lastLayer = self._model.layers 
-
-        shape = firstLayer.input_shape
+
```
lovelysandlonelys, line 71: Would you be able to explain this change to me?

lovelysandlonelys (changes requested):
I think that the versions for all the new files should just be 1.0.0. rather than 1.0.1 (which is kinda a nasty change sorry). Otherwise everything looks good to me 👍🏻

src/ModelExecution/modelWrapper.py
lovelysandlonelys: OHhh wait I just read the title, this is the bug fix for the tensorflow update?

lovelysandlonelys (approved):
LGTM 👍🏻

src/ModelExecution/modelWrapper.py
```diff
@@ -67,11 +68,12 @@ def __shape_data(self, inputs: array) -> None:
...
-
-        #Get only first and last layers
-        firstLayer, *_, lastLayer = self._model.layers 
-
-        shape = firstLayer.input_shape
+      
+        
+        shape = self._model.input_shape
+        
+        log(f'input shape {shape}')
```
bestrada33, line 75: Should this still be here?

data/dspec/Inundation/July/ar_inundation_july_12h.json
```diff
@@ -0,0 +1,63 @@
+{
+    "modelName": "ar_inundation_july_12h",
```
bestrada33, line 2: I see that you switched from using the full months "apil" to just the first 3 letters "apr". Can you do that for june and july (so "jun" & "jul") too? Just so they can all have the same format. (I need this for the script I'm making for CBOCP that automatically switches the model based on the month.)

data/dspec/Inundation/September/ar_inundation_sept_12h.json
```diff
@@ -0,0 +1,63 @@
+{
+    "modelName": "ar_inundation_sept_12h",
```
bestrada33, line 2: Same thing here as above. Can you use "sep" instead of "sept"?

data/cron/semaphore.cron
```diff
@@ -20,18 +20,18 @@
...
-5 * * * * mkdir -p ./logs/$(date "+\%Y")/$(date "+\%m")/Inundation/September && docker exec semaphore-core python3 src/semaphoreRunner.py -d ./Inundation/September/ar_inundation_sept_24h.json >> ./logs/$(date "+\%Y")/$(date "+\%m")/Inundation/September/ar_inundation_september_24h.log 2>> ./logs/$(date "+\%Y")/$(date "+\%m")/Inundation/September/ar_inundation_september_24h_CRON.log 
-5 * * * * mkdir -p ./logs/$(date "+\%Y")/$(date "+\%m")/Inundation/September && docker exec semaphore-core python3 src/semaphoreRunner.py -d ./Inundation/September/ar_inundation_sept_12h.json >> ./logs/$(date "+\%Y")/$(date "+\%m")/Inundation/September/ar_inundation_september_12h.log 2>> ./logs/$(date "+\%Y")/$(date "+\%m")/Inundation/September/ar_inundation_september_12h_CRON.log 
+5 * * * * mkdir -p ./logs/$(date "+\%Y")/$(date "+\%m")/Inundation/September && docker exec semaphore-core python3 src/semaphoreRunner.py -d ./Inundation/September/ar_inundation_sep_48h.json >> ./logs/$(date "+\%Y")/$(date "+\%m")/Inundation/September/ar_inundation_sepember_48h.log 2>> ./logs/$(date "+\%Y")/$(date "+\%m")/Inundation/September/ar_inundation_sepember_48h_CRON.log
```
bestrada33, line 32: September is spelt wrong in all of these. Also, any reason for using the full month name for the cron logs?

lovelysandlonelys (approved):

PR #307, opened 2024-05-09
Refactored and created the logic for interpolation

abeasley1722:
### Changes:

Refactor(Pairprogrammed with Savannah):

- Added maxGapDistance to the dependent series object
-  Refactored dspec parser to add the maxGapDistance 
- Refactored input gatherer 
- Refactored series provider

Added Interpolation Logic(Pairprogrammed with Beto):

In SeriesProvider:
- Created the interpolation function
- Created the check_gap_distance function
- Created the fill_in_date_gaps function


src/SeriesProvider/SeriesProvider.py
```diff
@@ -79,9 +79,134 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+        Returns:
+            Series : The merged result of the DB and DI queries
+        """
+        
+         merged_df = pd.DataFrame([input.__dict__ for input in validated_merged_result.data])
+         
+         merged_df.set_index("timeVerified", inplace=True)
+         
+         #Fill in the missing predictions and timestamps with NaNs
+         filled_merged_df = fill_in_date_gaps(merged_df,timeDescription.fromDateTime, timeDescription.toDateTime, timeDescription.interval,)
```
bestrada33, line 108: remove the comma after timeDescription.interval

src/SeriesProvider/SeriesProvider.py
```diff
@@ -79,9 +79,134 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+         #interpolate with time series interpolation
+         largerThanMaxGapDistance, rows_with_nan = check_gap_distance(filled_merged_df, seriesDescription.maxGapDistance, timeDescription.interval)
```
bestrada33, line 110: I understand the intent, but I don't think we need to return the rows_with_nan. It makes it look messy and there's no reason to print them below either.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -79,9 +79,134 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+         if largerThanMaxGapDistance:
+             log(f'''Interpolation error,
+                    Reason: There are gaps between data that is larger than the maxGapDistance limit.\n 
+                    {seriesDescription} \n 
+                    {timeDescription} \n 
+                    maxGapDistance: {seriesDescription.maxGapDistance}
+                    Rows with NaN only: {rows_with_nan}
```
bestrada33, line 117: What I mentioned above. Can remove this

src/SeriesProvider/SeriesProvider.py
bestrada33, line 108: Also, found an issue with this. Forgot to come back to it. The fill_in_date_gaps will add the missing timeVerified index column and make the rest of the columns NaN. So, we need to fill in the NaNs for the rest of the columns. (i.e dataUnit, timeGenerated, longitude, latitude.) dataValue will be interpolated later.

src/SeriesProvider/SeriesProvider.py
abeasley1722: Dangg. Will do.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -79,9 +79,134 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+                nan_gap_size += 1
+
+            # If the NaN gap is greater than maxGapDistance, return True
+            if(nan_gap_size > maxGapDistance):
+                return True
+
+            previous_date = current_date
+
+        # If no NaN gap was greater than maxGapDistance
+        return False, rows_with_nan
```
bestrada33, line 185: Just return False. (Also worth mentioning that False is "good" in this case, since it means that the interpolation was successful.)

src/SeriesProvider/SeriesProvider.py
```diff
@@ -79,9 +79,134 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+                 longitude=row["longitude"],
+                 latitude=row["latitude"]
+             ))  
+         validated_merged_result.data = inputs
+         validated_merged_result.isComplete = True
+         return validated_merged_result
+     
+    def  check_gap_distance(df: pd.DataFrame, maxGapDistance: timedelta, interval: timedelta):
+        """This method will remove all rows with successfull queries leaving just the NaN values. It will then check to make 
+        sure that there isn't a distance between two NAN values that is larger than the maxGapDistance.
```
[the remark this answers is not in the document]
abeasley1722: Thankss

src/SeriesProvider/SeriesProvider.py
```diff
@@ -70,7 +70,7 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
         if(validated_merged_result.isComplete):
             return validated_merged_result
         
-        # If neither were valid then we log this occurrence. We still return the best result we have, which is the merged result 
+        # If neither were valid then we log the occurence of missing values and check for interpolation. Then we return the interpolated result.
         log(f'''Request input failure, 
                     Reason: {validated_merged_result.nonCompleteReason} \n 
                     {seriesDescription} \n
```
matdenkas, line 76: I dont think we should be logging here. This is a silent action. If we had better logging we could make this verbose, and in the future we might want to have tracking on predictions to see if interpolation was used. However, this makes it look like the input was failed, and there is no mention of if interpolation happened. Personally I would only log a failure if we can't interpolate. But if you do want to log here, then it should mention if interpolation happened or not.

matdenkas (changes requested):
Very nice work! I dont like how dirty our SeriesDesc object is getting, but thats a problem for another day. This was a lot of code and I am thankful to be part of a team that took care of it for me 🫡

src/SeriesProvider/SeriesProvider.py
```diff
@@ -79,9 +79,133 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+        
+        if seriesDescription.maxGapDistance is not None:
+            
+            interpolated_merged_results = interpolate_series(seriesDescription, timeDescription, validated_merged_result)
+            return interpolated_merged_results
+        
         return validated_merged_result
+      
+    
+    def interpolate_series(seriesDescription: SeriesDescription, timeDescription: TimeDescription, validated_merged_result: Series) -> Series:
```
lovelysandlonelys, line 91: We have to pass self to use internal function so interpolate_series, fill_in_date_gaps, and check_gap_distance have to be private I think. (I can fix this and push it!)

src/SeriesProvider/SeriesProvider.py
```diff
@@ -79,9 +79,133 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+            timeDescription (TimeDescription): Describes the date time properties of the dataset
+            validated_merged_result (Series): The merged result of the DB and DI queries
+
+        Returns:
+            Series : The merged result of the DB and DI queries
+        """
+        
+         merged_df = pd.DataFrame([input.__dict__ for input in validated_merged_result.data])
+         
+         merged_df.set_index("timeVerified", inplace=True)
```
lovelysandlonelys, line 105: there is smth going on here but I'm not sure what it is: 
File "/app/src/SeriesProvider/SeriesProvider.py", line 105, in __interpolate_series
    merged_df.set_index("timeVerified", inplace=True)
  File "/usr/local/lib/python3.11/dist-packages/pandas/util/_decorators.py", line 331, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/pandas/core/frame.py", line 6012, in set_index
    raise KeyError(f"None of {missing} are in the columns")
KeyError: "None of ['timeVerified'] are in the columns"
05/09/24 18:57:10: Semaphore fin!

lovelysandlonelys (changes requested):

src/SeriesProvider/SeriesProvider.py
abeasley1722: This log was here before our changes. The only log we added for this task is on line 112. I see what you mean. Now that there is interpolation its probably not necessary anymore.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -79,9 +79,133 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+                 timeGenerated=row["timeGenerated"],
+                 timeVerified=row["timeVerified"],
+                 longitude=row["longitude"],
+                 latitude=row["latitude"]
+             ))  
+         validated_merged_result.data = inputs
+         validated_merged_result.isComplete = True
+         return validated_merged_result
+     
+    def  __check_gap_distance(df: pd.DataFrame, maxGapDistance: timedelta, interval: timedelta) -> bool:
```
bestrada33, line 136: This also needs `self` added as an argument.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -79,9 +79,133 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+            # If the NaN gap is greater than maxGapDistance, return True
+            if(nan_gap_size > maxGapDistance):
+                return True
+
+            previous_date = current_date
+
+        # If no NaN gap was greater than maxGapDistance
+        return False
+             
+    def __fill_in_date_gaps(df : pd.DataFrame , start_date : datetime , end_date : datetime , interval : timedelta ) -> pd.DataFrame:
```
bestrada33, line 186: This also needs `self` added as an argument.

src/SeriesProvider/SeriesProvider.py
bestrada33: Nice catch and thank you for fixing it!

lovelysandlonelys:
Notes From Meeting: 
- change seconds value of maxGapDistance to correspond to points since that's what we use to describe series in dspec and might make the method more generalized
- passing a lot to the interpolation method, maybe should only be passing the series
- should the method return the full series (create a new series with all the gaps filled) or return only the values that were interpolated
- To change the isComplete value we should be passing it to the validate series function NOT marking complete inside the interpolation logic
     - interpolation function should ONLY do the interpolation
- maybe change interpolation object in dspec to mandatory but if not there just sets interpolation to false?
     - true/false
     - interpolation method
     - gapDistance

src/SeriesProvider/SeriesProvider.py
bestrada33: Just for reference, but this is what I am referring to.
image

See how the rest of the columns are set to NaN or NaT. They need to be filled in with either forward fill or backward fill.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -70,18 +70,158 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+        interpolation_results = self.__interpolate_series(validated_merged_result)
+        validated_interpolation_results = self.__generate_resulting_series(seriesDescription, timeDescription, interpolation_results.data)
+        
+        if(validated_interpolation_results.isComplete):
+            return validated_interpolation_results
+        
+        return validated_interpolation_results
+      
+    
+    def __interpolate_series(self, inSeries: Series) -> Series:
```
bestrada33, line 84: Somewhere, we need to check if there is only one value being requested. If there is only one value, then we should not attempt to interpolate.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -70,13 +70,15 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
         
-        # If neither were valid then we attempt to interpolate
-        log(f'Init Interpolation from {seriesDescription}|{timeDescription}')
-        interpolation_results = self.__interpolate_series(validated_merged_result)
-        validated_interpolation_results = self.__generate_resulting_series(seriesDescription, timeDescription, interpolation_results.data)
-        
-        if(validated_interpolation_results.isComplete):
-            return validated_interpolation_results
+        #If there is only one Input (Only one data point) then we do not interpolate
+        if(len(validated_merged_result.data) > 1):
```
bestrada33, line 74: This won't work since we need validated_interpolation_results to be created. Will move inside the interpolate_series function

data/dspec/test_dspec-2-0.json
```diff
@@ -27,7 +27,10 @@
             "unit": "meter",
             "interval": 3600,
             "range": [0, 0],
+            "interpolationParameters": {},
```
matdenkas, line 30: This will cause issues. The get method in the parser will only apply the default if nothing is found. Technically there is something here though iirc. From my understanding this is like if we are expecting a sting and we get none/null or an empty string ('')

matdenkas: You can just delete the whole interpolation Parameters line

src/SeriesProvider/SeriesProvider.py
```diff
@@ -70,18 +70,188 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
-                    Series Storage Result: {series_storage_results.data} \n 
-                    Data Ingestion Result: {None if data_ingestion_results == None else data_ingestion_results.data} \n
-                    Merged Result: {validated_merged_result.data}
-            ''')
-        return validated_merged_result
+                ''')
+             return inSeries
 
-            
+         if interpolationParameters is None:
```
matdenkas, line 103: This should hard fail, it should raise an exception.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -70,18 +70,188 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+             log(f'''Interpolation error,
+                    Reason: Interpolation Parameters not defined.\n 
+                    {seriesDescription} \n 
+                    {timeDescription} \n 
+                ''')
+             return inSeries
+         
+         method = interpolationParameters.get("method")
+         limit = interpolationParameters.get("limit")
+         if method is None or limit is None:
```
matdenkas, line 113: Again, this should hard fail

src/SeriesProvider/SeriesProvider.py
```diff
@@ -70,18 +70,188 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+         
+         # Add the missing timeVerified datetimes and fill the remaining columns with NaNs/NaTs
+         filled_input_df = self.__fill_in_date_gaps(input_df, timeDescription.fromDateTime, timeDescription.toDateTime, timeDescription.interval)
+         
+         # Rename the index back to 'timeVerified' after filling in date gaps
+         filled_input_df.index = filled_input_df.index.rename('timeVerified')
+
+         largerThanLimit = self.__check_gap_distance(filled_input_df, limit, timeDescription.interval)
+         if largerThanLimit:
+             log(f'''Interpolation error,
```
matdenkas, line 135: This should be concatted to the non_complete reason on the series. Basically helping to log what went wrong with the series in the object itself.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -70,18 +70,188 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+        nan_gap_size = 1
+        for current_date in rows_with_nan.index:
+
+            # If first date, set the previous date to the current date and then continue
+            if previous_date is None:
+
+                previous_date = current_date
+
+                continue
+
```
matdenkas, line 211: If you want to remove this extra condition just for the first time, you can just set previous_date to the first index's date

src/SeriesProvider/SeriesProvider.py
```diff
@@ -70,18 +70,188 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
+        It will then check to see if there is a group of consecutive NaNs that are larger than the limit.
+
+        Args:
+            df (pd.DataFrame): The DataFrame of data
+            limit (timedelta): The max gap distance the researcher allows for their model
+            interval (timedelta): The time step separating the data points in order
+
+        Returns:
+            bool: Determines if a gap is larger than the limit
+        """
```
matdenkas, line 194: Amazing work! This is an impressively well written algo, I really like it!!!

src/SeriesProvider/SeriesProvider.py
```diff
@@ -173,7 +343,7 @@ def __generate_resulting_series(self, seriesDescription: SeriesDescription, time
 
             # If a result was missing, an input was not mapped to that key. Thus its mapped to None. We remove the whole k:v pair as to not pollute the results with None.
             datetimeDict = {k : v for k, v in datetimeDict.items() if v is not None}
-
+            #Taking out the entire pair including the timestamp may make it difficult to interpolate
```
matdenkas, line 346: Is this supposed to still be here? It doesn't make a lot of sense to me

src/tests/UnitTests/test_Interpolation.py
```diff
@@ -0,0 +1,88 @@
...
+
+    inSeries = Series(description = seriesDescription, isComplete = False, timeDescription = timeDescription)
+
+    inSeries.data = inputs
+
+    outSeries = seriesProvider._SeriesProvider__interpolate_series(inSeries)
+
+    actual_length_of_data = len(outSeries.data)
+
+    assert actual_length_of_data == expected_length_of_data
```
matdenkas, line 88: Really really solid test. I love how you parametrized it. As a critique watch your edge cases. A empty series and a series that doesn't actually need interpolation. But seriously, this is already a really really good test.

matdenkas (changes requested):
This was so much code! Oh my god! Feel proud this work is really really solid!

src/SeriesProvider/SeriesProvider.py
bestrada33: <3

src/SeriesProvider/SeriesProvider.py
bestrada33: Not sure what this was. Removed.

PR #308, opened 2024-05-09
290 create semaphore ingestion class

matdenkas:
```python
import sys
sys.path.append('/app/src')

import pytest

from datetime import datetime, timedelta, time, date
from src.DataClasses import TimeDescription, SeriesDescription
from src.DataIngestion.IDataIngestion import data_ingestion_factory
from src.DataIngestion.DI_Classes.LIGHTHOUSE import LIGHTHOUSE
from dotenv import load_dotenv


load_dotenv()

sd = SeriesDescription("SEMAPHORE", "test AI|1.0.0|testSeries", 'packChan', "test_datum") 
td = TimeDescription(datetime.combine(date(2024, 4, 28), time(0, 0)), datetime.combine(date(2024, 4, 30), time(12, 0)), timedelta(seconds=3600))

lighthouse = data_ingestion_factory(sd)
result = lighthouse.ingest_series(sd, td)

print(result)
print(result.data)
```

Code provided for testing. I didnt make this into a test as that would rely on DB data.

Note that the information about the model has to be packed together into the series name. This was the best way I could think to do it, lmk if you think of a better way

### DIFF
--- a/src/DataIngestion/DI_Classes/SEMAPHORE.py
+++ b/src/DataIngestion/DI_Classes/SEMAPHORE.py
@@ -54,11 +54,11 @@ class SEMAPHORE(IDataIngestion):
 
         # Repack the data with the original series description and return
         return_series = Series(seriesDescription, True, timeDescription)
-        return_series.data = self.__convert_input_to_output(result.data) # Cast outputs to inputs
+        return_series.data = self.__convert_output_to_input(result.data) # Cast outputs to inputs
         return return_series
     
 
-    def __convert_input_to_output(self, outputs: list[Output]) -> list[Input]:
+    def __convert_output_to_input(self, outputs: list[Output]) -> list[Input]:
         """A simple method to cast and output object into an input object"""
         inputs = []
         for output in outputs:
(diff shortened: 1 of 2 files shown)


lovelysandlonelys (approved):
Makes sense to me and test code provided ran 👍🏻

src/DataIngestion/DI_Classes/SEMAPHORE.py
```diff
@@ -0,0 +1,79 @@
...
+        # Fetch data through the ORM
+        result = self.series_storage.select_output(ssd, timeDescription)
+
+        # Repack the data with the original series description and return
+        return_series = Series(seriesDescription, True, timeDescription)
+        return_series.data = self.__convert_input_to_output(result.data) # Cast outputs to inputs
+        return return_series
+    
+
+    def __convert_input_to_output(self, outputs: list[Output]) -> list[Input]:
```
bestrada33, line 61: Shouldn't this be convert_output_to_input?

matdenkas: UUUUUUUUUUUUUUUUUUUUUUUUUUUGH

matdenkas: Fixed

bestrada33 (approved):
LGTM. I think the seriesName workaround is fine. I wish we could use the API, but as Florence said, it would be hard to have to switch between APIs on different servers. I do it for CBOCP, but it involves using an env variable to switch the host

PR #309, opened 2024-05-09
285 create the dywinds and the dxwinds post processing class

matdenkas:
Test command:
```bash
docker exec semaphore-core python3 -m pytest ./src/tests/UnitTests/test_ResolveVectorComponents.py
```

I renamed a thing to keep the class generalized:
[Computing vector components is called resolving](http://hyperphysics.phy-astr.gsu.edu/hbase/vect.html)

### DIFF
--- a/src/PostProcessing/PostProcessingClasses/ResolveVectorComponents.py
+++ b/src/PostProcessing/PostProcessingClasses/ResolveVectorComponents.py
@@ -56,7 +56,7 @@ class ResolveVectorComponents(IPostProcessing):
 
         # Unpack arguments from arg object
         args = postProcessCall.args
-        offset = args['offset']
+        offset = float(args['offset'])
         magnitude_series = preprocessedData[args['targetMagnitude_inKey']]
         direction_series = preprocessedData[args['targetDirection_inKey']]
 
@@ -68,11 +68,11 @@ class ResolveVectorComponents(IPostProcessing):
             direction = dir_input.dataValue
             magnitude = mag_input.dataValue
 
-            # Direction my be in degrees for calculation
+            # Direction must be in degrees for calculation
             if(dir_input.dataUnit != 'degrees'): direction = degrees(direction)
 
-            x_comp = magnitude * cos(radians(float(direction - offset)))
-            y_comp = magnitude * sin(radians(float(direction - offset)))
+            x_comp = magnitude * cos(radians(float(direction + offset)))
+            y_comp = magnitude * sin(radians(float(direction + offset)))
 
             # Magnitude contains the correct metadata from resulting series
             x_comps.append(Input(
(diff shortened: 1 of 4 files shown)


src/PostProcessing/PostProcessingClasses/ResolveVectorComponents.py
```diff
@@ -0,0 +1,116 @@
...
+                "targetMagnitude_inKey":"",
+                "targetDirection_inKey":"",
+                "x_comp_outKey": "", 
+                "y_comp_outKey": ""      
+            }
+        },
+
+    """
+    def post_process_data(self, preprocessedData: dict[str, Series], postProcessCall: PostProcessCall ) -> dict[str, Series]:
+        """Abstract method to define the post-processing operation.
```
lovelysandlonelys, line 47: Tis no longer and abstract method methinks

lovelysandlonelys (approved):
:O looks good to me (one documentation thing)
i like that in the postprocessing class description you included what the call would look like in the json!

src/PostProcessing/PostProcessingClasses/ResolveVectorComponents.py
bestrada33: What a sentence

matdenkas: Fixed! Ttytytytyty!

src/PostProcessing/PostProcessingClasses/ResolveVectorComponents.py
```diff
@@ -0,0 +1,116 @@
...
+        y_comps = []
+        for mag_input, dir_input in zip(magnitude_series.data, direction_series.data):
+
+            direction = dir_input.dataValue
+            magnitude = mag_input.dataValue
+
+            # Direction my be in degrees for calculation
+            if(dir_input.dataUnit != 'degrees'): direction = degrees(direction)
+
+            x_comp = magnitude * cos(radians(float(direction - offset)))
```
bestrada33, line 74: What if they want to add a rotation? Like +20 degrees instead of -20? Could we do (direction + offset) here and allow negative offsets?

src/PostProcessing/PostProcessingClasses/ResolveVectorComponents.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # Itterate through the magnitude and direction data, calculating components, and saving results in Input objects
+        x_comps = []
+        y_comps = []
+        for mag_input, dir_input in zip(magnitude_series.data, direction_series.data):
+
+            direction = dir_input.dataValue
+            magnitude = mag_input.dataValue
+
+            # Direction my be in degrees for calculation
```
bestrada33, line 71: may be? or must be?

PR #320, opened 2024-05-14
Fixing circular imports

matdenkas:
This is very much a patchwork fix. I went to a friend about this. She said that its likely systemic of me having used `from foo import *` Basically she thinks `import foo` should have been used instead. This helps break circular imports. This solution will work for now though.


lovelysandlonelys (approved):
we love imports 😢

PR #365, opened 2024-07-10
363 edit requirementstxt

abeasley1722:
This doesn't seem correct, but it is whats running on sherlock dev.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,5 +9,6 @@ requests>=2.31.0
 SQLAlchemy>=2.0.20
 tensorflow>=2.14.0
 psycopg2-binary==2.9.9
-pandas>-2.2.1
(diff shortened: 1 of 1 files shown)


lovelysandlonelys (approved):
if it's what running it's what's running ig

requirements.txt
```diff
@@ -9,5 +9,6 @@ requests>=2.31.0
 SQLAlchemy>=2.0.20
 tensorflow>=2.14.0
 psycopg2-binary==2.9.9
-pandas>-2.2.1
```
matdenkas, line 12: I think whoever wrote this wanted >= 2.2.1 THIS IS WHY WE USE SPACES >:(

LGTM though!

matdenkas (approved):

PR #370, opened 2024-07-12
HOHONU Data Ingestion Class

abeasley1722:


src/DataClasses.py
```diff
@@ -118,20 +120,24 @@ class SeriesDescription():
         :param limit: timedelta - The max gap distance allowed for interpolation
         :param dataDatum: str = None
         :param dataIntegrityDescription: DataIntegrityDescription = None - An object that contains the data integrity calls that should be handled.
+        :param is_output: bool = None - A bool that tells Semaphor whether this information will be used as an output or input
+        :param station_id: str = The stations ID
     """
-    def __init__(self, dataSource: str, dataSeries: str, dataLocation: str, dataDatum: str = None,  dataIntegrityDescription: DataIntegrityDescription = None, verificationOverride: int = None) -> None:
+    def __init__(self, dataSource: str, dataSeries: str, dataLocation: str, dataDatum: str = None,  dataIntegrityDescription: DataIntegrityDescription = None, verificationOverride: int = None, is_output: bool = None, station_id: str = None) -> None:
```
lovelysandlonelys, line 126: regarding including the station ID in the series description, I don't think that including something in the series description that's _specific_ to a certain injestion class is necessarily a good idea. 

I was under the impression that if an ingestion class needed a specific internal location code or ID then we stored that in the semaphore database alongside the data source name: 
image

it has been a while since I've looked at/worked with semaphore though so please please correct me if I'm wrong but I think that we should be utilizing our database and the ORM method below for specific station ID's 

image

src/DataIngestion/DI_Classes/HOHONU.py
```diff
@@ -0,0 +1,212 @@
...
+    def ingest_series(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> Series | None:
+
+        api_host = getenv('HOHONU_API_HOST')
+        api_authorization = getenv('HOHONU_API_AUTH')
+        
+        if not api_host or not api_authorization:
+            raise ValueError(f"API host or authorization is empty.")
+        
+        datum = seriesDescription.dataDatum
+        station_id = seriesDescription.station_id
```
lovelysandlonelys, line 36: so in line with my previous comment i think that rather than having station ID as a member of seriesDescription you would use a call to the ORM to find what hohonu calls the station that we need like this example below in the lighthouse ingestion class: 

image

src/DataIngestion/DI_Classes/HOHONU.py
```diff
@@ -0,0 +1,212 @@
...
+            
+            series.data = self.__convert_output_to_input(outputs)
+            
+        else:
+            series.data = outputs       
+    
+        return series
+    
+    def __corrects_interval(self, response_df: pd.DataFrame, timeDescription: TimeDescription):
+        """This script ensures the correct interval of datapoints is being used"""
```
lovelysandlonelys, line 85: description doesn't list parameters or returns

src/DataIngestion/DI_Classes/HOHONU.py
```diff
@@ -0,0 +1,212 @@
...
+    
+        # Resample the DataFrame to a different interval (e.g., 30 seconds)
+        new_interval = f"{timeDescription.interval.total_seconds()}S" # (e.g., '30S' for 30 seconds)
+        
+        # Ensures the correct interval is used 
+        resampled_df = response_df.resample(new_interval).interpolate('linear')
+        
+        return resampled_df
+        
+    def __convert_date_format(self, date):
```
lovelysandlonelys, line 100: function definition doesn't have the datatypes of passed parameter like other functions do

src/DataIngestion/DI_Classes/HOHONU.py
```diff
@@ -0,0 +1,212 @@
...
+        Args:
+            hohonu_df (pd.DataFrame): Datatframe containing hohonu station data
+
+        Returns:
+            hohonu_df (pd.DataFrame): Datatframe containing hohonu station data
+        """
+        
+        return date.strftime('%Y-%m-%d %H:%M:%S') 
+    
+    def __convert_dataframe_to_output(self,df: pd.DataFrame, timeDescription: TimeDescription) ->list[Output]:
```
lovelysandlonelys, line 112: no function description

src/tests/IntegrationTests/test_HOHONU.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+#test_LIGHTHOUSE.py
```
lovelysandlonelys, line 2: i haven't run the test mostly because of my other comments but the other aspects of the ingestion class seem good to me.

lovelysandlonelys (changes requested):

src/DataIngestion/DI_Classes/HOHONU.py
```diff
@@ -0,0 +1,188 @@
...
+        """Converts a dataframe 
+        :param date: A date time
+        """
+        df = df.reset_index()
+        inputs = []
+        for timestamp, value in zip(df["timestamp"], df["value"]):
+            dt = timestamp.to_pydatetime()
+            dataPoint = Input(
+                dataValue= value,
+                dataUnit= "feet",
```
matdenkas, line 115: This is an imperial unit, do we want to make an exception for this? We have a strong rule that everything in the DB should be metric!

src/DataIngestion/DI_Classes/HOHONU.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/local/bin/python
+#NDBC.py
```
matdenkas, line 2: NDBC -> HOHONU

src/DataClasses.py
```diff
@@ -97,17 +97,19 @@ class TimeDescription():
...
-    def __init__(self, fromDateTime: datetime, toDateTime: datetime, interval: timedelta = None) -> None:
+    def __init__(self, fromDateTime: datetime, toDateTime: datetime, interval: timedelta = None, leadtime: timedelta = None) -> None:
         self.fromDateTime = fromDateTime
         self.toDateTime = toDateTime
         self.interval = interval
+        self.leadtime = leadtime
 
     def __str__(self) -> str:
-        return f'\n[TimeDescription] -> fromDateTime: {self.fromDateTime}, toDateTime: {self.toDateTime}. interval: {self.interval}'
+        return f'\n[TimeDescription] -> fromDateTime: {self.fromDateTime}, toDateTime: {self.toDateTime}, interval: {self.interval}, leadtime: {self.leadtime}'
```
matdenkas, line 109: Why are we adding this to this data structure? Maybe I don't know whats going on, but lead time is a attribute of a model, not a span of time. The dime description should describe the datetime properties of a dataset, meaning the span of time that the data is in, in what interval.

matdenkas (changes requested):
Can you email us the HOHONU api keys so that we can test on our machines? ty! Im not keyed in yet on whats going on with this Hohonu stuff so if anything im asking is completely ignorant my b!

src/DataIngestion/DI_Classes/HOHONU.py
```diff
@@ -0,0 +1,188 @@
...
+        df = df.reset_index()
+        inputs = []
+        for timestamp, value in zip(df["timestamp"], df["value"]):
+            dt = timestamp.to_pydatetime()
+            dataPoint = Input(
+                dataValue= value,
+                dataUnit= "feet",
+                timeVerified= dt + timeDescription.leadtime,
+                timeGenerated= dt,
+                leadTime= timeDescription.leadtime
```
lovelysandlonelys, line 118: I think this is the only place where you use the leadtime? to calculate a timeVerified? 

I'm about to expose the extent to which I do not remember what is going on with the magnolia models but are we getting predictions or are we getting actual datapoints? From what you did here I'm assuming it's predictions. Does hohonu truly not give any information about when their predictions are for in the returned data? It looks like in the other ingestion classes either the time generated is used for both, or (in ndfd) the time generated is set to "None" so....... probably something we should talk about, potentially?

PR #378, opened 2024-07-31
changed the lighthouse url

abeasley1722:
We are required to use the encrypted version of http in order for lighthouse API requests to work as expected. This will allow magnolia, thermal refuge, and mlp-op to retrieve their needed data again.


lovelysandlonelys (approved):
in line with the message you forwarded us from niall

PR #384, opened 2024-08-07
SeriesProvider now saves the series

abeasley1722:


src/SeriesProvider/SeriesProvider.py
```diff
@@ -28,7 +28,7 @@ class SeriesProvider():
     def __init__(self) -> None:
         self.seriesStorage = series_storage_factory()
     
-    def save_series(self, series: Series) -> Series:
+    def save_output_series(self, series: Series) -> Series:
```
lovelysandlonelys, line 31: why have this function if we don't use it anywhere?

src/SeriesProvider/SeriesProvider.py
```diff
@@ -37,12 +37,27 @@ def save_series(self, series: Series) -> Series:
...
         return returningSeries
     
+    def save_input_series(self, series: Series):
+        """Passes a series to Series Storage to be stored.
+            :param series - The series to store.
+            :returns series - A series containing only the stored values.
+        """
+        returningSeries = Series(series.description, True)
+
+        if (type(series.description) == SemaphoreSeriesDescription): #Check and make sure this is actually something with the proper description to be inserted
```
lovelysandlonelys, line 53: I totally forgot about this wehn we were talking about it, but we need to check that the series wasn't passed by the semaphore ingestion class because thats the one class we don't want to save in the database ever since it pulls it's data from the database.

lovelysandlonelys (changes requested):

lovelysandlonelys (approved):

PR #404, opened 2024-08-21
2.4 database migration for HOHONU

abeasley1722:


src/tests/IntegrationTests/test_SeriesProvider.py
```diff
@@ -27,10 +27,24 @@ def test_save_series():
     load_dotenv()
     seriesProvider = SeriesProvider()
     
-    seriesDescription = SeriesDescription('apple', 'pear', 'grape')
+    inputs = [
```
matdenkas, line 30: This test can be improved. You should test edge cases and an expected case. So not only inserting a proper series, but one thats expected to fail, and probably also an empty series. Tests also should not assert true, you should assert the expected and the actual that way not only will the test fail if an exception occurred but also if non expected behavior is happening.

tools/DatabaseMigration/2_4/init_data/hohonuDataDatum.csv
```diff
@@ -0,0 +1,2 @@
+code,displayName,notes
+D2W_FEET,Distance to Water in Feet, Distance to Water in Feet
```
matdenkas, line 2: I dont think we should be inserting imperial data into our database. We have a hard rule that its only metric and we convert in and out of metric as needed. If this was discussed please ignore me.

That being said this being the datum, datums are unit independent. So it shouldn't mention feet at all. The datum would be D2W and the unit code would be feet (though hopefully meter)

matdenkas (changes requested):
I need the Hohonu API keys to test. Again im a bit keyed out with the hohonu stuff so if im being ignorant my b

lovelysandlonelys (commented):
aside from the FEET everything _looks_ right, I also need the api keys for testing

PR #406, opened 2024-08-23
2.4 Database Migration for HOHONU

abeasley1722:


abeasley1722:
Matthew's Notes from the last Pull Request: 

I dont think we should be inserting imperial data into our database. We have a hard rule that its only metric and we convert in and out of metric as needed. If this was discussed please ignore me.

That being said this being the datum, datums are unit independent. So it shouldn't mention feet at all. The datum would be D2W and the unit code would be feet (though hopefully meter)

tools/DatabaseMigration/2_4/2_4_DatabaseMigration.py
```diff
@@ -0,0 +1,97 @@
+# -*- coding: utf-8 -*-
+#2_4_DatabaseMigration.py
+#----------------------------------
+# Created By: Anointiyae Beasley
+# Created Date: 08/21/2024
+# Version 1.0
+#----------------------------------
+"""This is a database migration script that will initialize version
+    2.4 of the database. The change from version 2.3 to 2.4 includes a new dataMapping, dataSource, and dataDatum for HOHONU.
```
lovelysandlonelys, line 9: It looks like now you're only including a dataMapping and a dataSource, so a small documentation update. Make sense to take out the dataDatum for Hohonu since we already added d2w in the 2.3 migration.

tools/DatabaseMigration/target_version.json
```diff
@@ -1,4 +1,4 @@
 {
-    "Target Version" : "2.3",
-    "Description" : "Adding one row to the datum reference table for the Magnolia model."
+    "Target Version" : "2.4",
+    "Description" : "Adding a source, mapping, and datum for HOHONU."
```
lovelysandlonelys, line 3: no datum

lovelysandlonelys (changes requested):
Two very very small documentation issues, everything else looks great and i tested rollback and migration on my machine from 2.3 to 2.4 and 2.4 to 2.3 and everything looks perfect.

PR #407, opened 2024-08-23
362-create-the-hohonu-ingestion-class

abeasley1722:


src/DataIngestion/DI_Classes/HOHONU.py
```diff
@@ -0,0 +1,188 @@
...
+        :param date: A date time
+        """
+        df = df.reset_index()
+        inputs = []
+        for timestamp, value in zip(df["timestamp"], df["value"]):
+            dt = timestamp.to_pydatetime()
+            dataPoint = Input(
+                dataValue= value,
+                dataUnit= "feet",
+                timeVerified= dt + timeDescription.leadtime,
```
lovelysandlonelys, line 116: image

timeDescription.leadtime doesn't exist as a part of the time description anymore, so I think here you would have to just do dt?

src/tests/IntegrationTests/test_HOHONU.py
```diff
@@ -0,0 +1,40 @@
...
+#Import
+from src.DataIngestion.IDataIngestion import data_ingestion_factory
+from src.DataClasses import SeriesDescription,TimeDescription,Input
+import pytest
+from datetime import datetime, timedelta
+from dotenv import load_dotenv
+import pandas as  pd
+
+@pytest.mark.parametrize("seriesDescription, timeDescription, expected_output", [
+    (SeriesDescription(dataSource='HOHONU', dataSeries='dWl', dataLocation='MagnoliaBeach', dataDatum= 'D2W_FEET', is_output=False), TimeDescription(fromDateTime=(datetime.now() - timedelta(hours=10)),toDateTime= (datetime.now()- timedelta(hours= 8) ), interval= timedelta(seconds = 3600), leadtime=timedelta(seconds = 43200)), 2),
```
lovelysandlonelys, line 22: is_output is no longer a part of the series description, you also still have leadtime as a part of the time description despite taking it out of the data class itself. 

image

lovelysandlonelys (changes requested):

src/DataIngestion/DI_Classes/HOHONU.py
```diff
@@ -0,0 +1,188 @@
...
+        """Converts a dataframe 
+        :param date: A date time
+        """
+        df = df.reset_index()
+        inputs = []
+        for timestamp, value in zip(df["timestamp"], df["value"]):
+            dt = timestamp.to_pydatetime()
+            dataPoint = Input(
+                dataValue= value,
+                dataUnit= "feet",
```
matdenkas, line 115: We shouldn't be putting imperial units into the DB. Technically we can but we are supposed to convert to metric and then convert back out when we use it else where. There is already a system in Semaphore to do this. LMK if you need help figuring it out.

matdenkas (changes requested):
LMK if you wanna talk about this. IDK if this has been talked over with people

PR #408, opened 2024-08-23
chartJson for Magnolia Beach

abeasley1722:


lovelysandlonelys (approved):
it looks good to meee, there isn't really a way to test this yet right? because we don't have a chart dspec parser yet or anything?

PR #420, opened 2024-08-27
Improved NDFD logging

matdenkas:
This PR adds a bit of code to correctly Identify the server side error that NDFD can return. It prevents semaphore from raising an internal error and logs it better.


lovelysandlonelys (approved):
image
Adding information about the type of curl information NDFD is throwing confirmed.

abeasley1722 (approved):
LGTM. This will help narrow down the many errors ndfd loves to give.

PR #427, opened 2024-09-04
Series provider refector w/ tests

matdenkas:
Okay, this started with a bug, that if an ingestion class returned an empty series that Series Provider would try and insert it into the DB causing a weird DB error.

In isolation that's a very simple bug fix, but this is the third bug ive fixed in series provider THIS WEEK. The fact is it had some issues. There were some architecture issues with how data was being handled and inserted and it was just incredibly unreadable and had awful tests.

Thus I refactored it to improve its readability and reworked how series provider sends data to get inserted. I wrote a mock data ingestion class and a mock series storage class and reworked the unit testing for series provider.

To test this (and please run through this with a fine tooth comb less we be back here next week to fix another bug):

1. `docker exec semaphore-core python3 -m pytest` To run the new tests
2. `docker exec semaphore-core python3 src/semaphoreRunner.py -d TestModels/test_dspec.json` run this twice, first to see its hitting both ingestion and then again to see its just getting the data from the DB

If you have comments on documentation or how to make this class more readable i'm all ears!!

### DIFF
--- a/src/DataIngestion/DI_Classes/TEST_DI.py
+++ b/src/DataIngestion/DI_Classes/TEST_DI.py
@@ -20,7 +20,6 @@ ERROR_RAISE - Actually will raise an exception for you to catch
 # 
 #
 #Input
-from SeriesStorage.ISeriesStorage import series_storage_factory
 from DataClasses import Series, SeriesDescription, Input, TimeDescription
 from DataIngestion.IDataIngestion import IDataIngestion
 
--- a/src/DataIngestion/DI_Classes/TEST_DI.py
+++ b/src/DataIngestion/DI_Classes/TEST_DI.py
@@ -56,22 +56,17 @@ class TEST_DI(IDataIngestion):
                     raise Exception(f'In testing DI class {amount_of_missing} from a series that only has {len(series.data)} values!')
                 
                 series.data = self.__generate_data_with_missing(series.data, amount_of_missing)
-                print(series.data)
                 return series
 
 
     def __generate_data_with_missing(self, data: list[Input], amount_of_missing: int) -> list[Input]:
         missing_idx = random.sample(range(0, len(data)), amount_of_missing)
-        print(missing_idx)
         new_data = []
-        print('before', data)
         for idx, data in enumerate(data):
             if idx in missing_idx:
-                print(f'Missing idx {idx} skipped!')
                 continue
             else:
                 new_data.append(data)
-        print('after', new_data)
         return new_data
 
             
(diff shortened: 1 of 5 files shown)


src/SeriesProvider/SeriesProvider.py
```diff
@@ -117,73 +92,100 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
+        series_storage_results = self.seriesStorage.select_input(seriesDescription, timeDescription)
+        validated_series_storage_results = self.__generate_resulting_series(seriesDescription, timeDescription, series_storage_results.data)
+        return validated_series_storage_results, series_storage_results
+        
+
+    def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> tuple[Series | None, Series | None]:
+        """ Handles the process of getting requested data from series storage, validating it, and returning both validated and raw results 
+        :param seriesDescription: SeriesDescription - The semantic description of request
+        :param timeDescription: TimeDescription - The temporal description of the request
+        :returns tuple[Series, Series] | None
```
matdenkas, line 116: Should be `:returns tuple[Series | None, Series | None]`

src/DataIngestion/DI_Classes/TEST_DI.py
```diff
@@ -0,0 +1,110 @@
...
+                    raise Exception(f'In testing DI class {amount_of_missing} from a series that only has {len(series.data)} values!')
+                
+                series.data = self.__generate_data_with_missing(series.data, amount_of_missing)
+                return series
+
+
+    def __generate_data_with_missing(self, data: list[Input], amount_of_missing: int) -> list[Input]:
+        missing_idx = random.sample(range(0, len(data)), amount_of_missing)
+        new_data = []
+        for idx, data in enumerate(data):
```
abeasley1722, line 65: Probably simple code, but I didn't know the "enumerate" function existed and that makes certain issues where index counting is required way easier.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -23,6 +23,7 @@
```
abeasley1722, line 24: datetime import isn't used

src/SeriesProvider/SeriesProvider.py
```diff
@@ -117,73 +92,100 @@ def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
...
-        # Construct a dictionary for the db results
-        database_results = { input.timeVerified : input for input in DBList }
-
-        # If there are data ingestion results construct a dictionary for that too
-        if DIList != None:
-            ingestion_results = { input.timeVerified : input for input in DIList}
+        # Construct a dictionary for the first results
+        first_results = { input.timeVerified : input for input in first}
 
+        if second != None:# If there are second results construct a dictionary for that too
```
abeasley1722, line 175: Picky, but might be better to use "is not"

matdenkas: Fixed, ty!

PR #429, opened 2024-09-12
Logging NDBC Error

lovelysandlonelys:
**Background:** 
For a period of time from late July to mid August semaphore experienced an 'index out of range' error. Because the error has only been observed within that time frame no further investigation was done, however it seemed like a good idea to add a small logging statement for when NDBC returns no data in order to provide future debuggers with more information/context. 

I just did a simple statement and added a comment about why it's likely an error on NDBC's side. Very open to changes that align the log more with what is appropriate to log. 

**To Test:** 
Run a model that uses NDBC to make sure the statement doesn't trigger when data is returned: 
`docker exec semaphore-core python3 src/semaphoreRunner.py -d Inundation/August/ar_inundation_aug_12h.json`

Run a model from the time period the error was observed to ensure log statement executes: 
`docker exec semaphore-core python3 src/semaphoreRunner.py -d Inundation/August/ar_inundation_aug_12h.json -p '202408220101'`

### DIFF
--- a/src/DataIngestion/DI_Classes/NDBC.py
+++ b/src/DataIngestion/DI_Classes/NDBC.py
@@ -162,7 +162,7 @@ class NDBC(IDataIngestion):
         full_series_inputs = self.__get_NDBC(seriesDescription, timeDescription).data # request the data
 
         if not full_series_inputs:  # check that data has been provided
-            log(f"Error: NDBC returned no data! Likely an NDBC outage.")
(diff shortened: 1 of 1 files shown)


matdenkas (approved):
LGTM, follows our previous established conventions! Lets hope we never see it again!

abeasley1722 (approved):
LGTM!

PR #437, opened 2024-09-13
Updating MLP-OP to DSPEC 2.0

matdenkas:
Updating MLP-OP dspec to 2.0. 

To check this PR... its gonna suck. I wrote a checklist you can find here:
https://docs.google.com/spreadsheets/d/1eV9fRGkZnrJYPFxOCErO9p_oEVL2Mu1whJM-uyYW-30/edit?usp=sharing

Besides that its just comparing the old and new version determining no mistakes were made in the update. Sorry 😭

### DIFF
--- a/data/dspec/MLP-OP.json
+++ b/data/dspec/MLP-OP.json
@@ -3,7 +3,7 @@
     "modelName": "MLP-OP",
     "modelVersion": "2.0.0",
     "author": "Christian Duff",
-    "modelFileName": "MLP-OP",
+    "modelFileName": "./MLP-OP",
     "timingInfo":{
         "active": true,
         "offset": 600,
@@ -26,7 +26,7 @@
             "unit": "celsius",
             "interval": 3600,
             "range": [0, -23],
-            "outKey": "NPSBI_DWL_24"
+            "outKey": "NPSBI_DWT_24"
         },
         { 
             "_name": "Air Temp",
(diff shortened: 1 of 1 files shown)


lovelysandlonelys (approved):
looks good to me and ran on my machine

data/dspec/MLP-OP.json
```diff
@@ -1,4 +1,5 @@
 {
+    "dspecVersion": "2.0",
     "modelName": "MLP-OP",
     "modelVersion": "2.0.0",
     "author": "Christian Duff",
```
lovelysandlonelys, line 5: in the other two requests you have ./filename but in this one you just have `"modelFileName": "MLP-OP",`

matdenkas: Both our valid paths, ./is more explicit. We use a pathing library so it understands what ./ means. Either is fine, im not sure which way would be better

lovelysandlonelys: i think that we have ./in most of the other files so maybe that just for consistency?

data/dspec/MLP-OP.json
```diff
@@ -16,36 +17,51 @@
...
             "_name": "Water Temp",
             "location": "NPSBI",
             "source": "LIGHTHOUSE",
             "series": "dWaterTmp",
             "unit": "celsius",
-            "type": "float",
             "interval": 3600,
-            "range": [0, -23]
+            "range": [0, -23],
+            "outKey": "NPSBI_DWL_24"
```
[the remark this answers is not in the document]
matdenkas: Fixed!

PR #472, opened 2024-09-24
277 write migration script for adding api user accounts

lovelysandlonelys:
Questions: 
- If  the database is ALREADY at the target version and we run relaunch the containers then the correct .yml file for the accounts to be used won't be there. However, we can't just change the .yml file because you can only use the accounts once they have been created. I wasn't sure how to solve this... maybe a detection in the actual migrate script that detects if the version is above 2.5 and replaces the yml? Or two different pull requests one with the db migration and one that follows with the changes to the yml file? That makes us not very backwards compatible even with the version that deploys the accounts though, and also technically isn't a database migration so I'm just not sure...
- Opening it up to the team in this pull request because I wasn't sure what the best way to go about this was :/

To Test: 
- Migrate the DB  to 2.5 `docker exec semaphore-core python3 tools/migrate_db.py`
- Check that the correct db connection strings are in the .yml `docker exec semaphore-core cat docker-compose.yml`
  - the connection strings should match what is in new-docker-compose.yml
- Rollback  the DB to 2.4 (change the target version) 
- check that the correct db connection strings are in the .yml

- Migrate to 2.5 from 2.5, the accounts are in the database but if you check the .yml file the containers won't be using their new accounts.


abeasley1722:
Will need to change the database version depending on whose code is ready first.

matdenkas (changes requested):
Good work as always. I am having some issues getting this to work. Here is what I did:

1. Checkout
2. Edit env from the dist:
    - Changed POSTGRES_USER and POSTGRES_PASSWORD to my currentd dbs user name and password
    - Changed CORE, API, and GENERAL to the user name and password I wanted them to use
3. Build and run
4. Ran supplied

I got an operational error when running the command: `sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server at "host.docker.internal" (192.168.65.254), port 5435 failed: FATAL:  password authentication failed for user "core_user"`

I could also see in the log other users were failing


Im not sure if something is wrong of if I did it wrong, could you please provide more detailed steps to update an existing db?

abeasley1722:
I am receiving this error. I do not get this error on the main branch.
Screenshot 2024-10-23 214039

abeasley1722 (changes requested):
I am receiving an error that I added in a previous comment. Overall, good work! I will try to help narrow down what could be causing this error.

lovelysandlonelys:
On ice for forever, lets close for now...

PR #476, opened 2024-09-29
test ai notifs driving me nuts

lovelysandlonelys:
The test ai notifications happen every hour becasue they're running into the error with NOAATANDC where the data isn't ready until five minutes after the hour so what if we offset that? Hopefully would reduce the number of notifications we get that don't actually mean anything.


matdenkas:
We disabled them

PR #477, opened 2024-09-30
Base of Frontend and its docker container

abeasley1722:
This is the base of the Vue app and frontend's docker container. The docker container will run a development server. This, of course, will be changed when we're ready for production. Run the containers like usual and confirm that localhost 8080 shows a blank white screen with "Semaphore" and a drop down menu. It looks very bland since I will be focusing on visuals last and logic first. (Other than the styes on the dropdown menu).
image


docker-compose.yml
```diff
@@ -82,9 +82,36 @@ services:
...
 
+  frontend:
+    container_name: semaphore-frontend
+    build:
+      context: .
+      dockerfile: Dockerfile.frontend
+    restart: always
+    env_file:
+      - .env
+    environment:
```
matdenkas, line 98: Do we need to expose all of these envs?

docker-compose.yml
```diff
@@ -82,9 +82,36 @@ services:
...
+      - DB_LOCATION_STRING=${DB_LOCATION_STRING}
+      - ISERIESSTORAGE_INSTANCE=${ISERIESSTORAGE_INSTANCE}
+      - DSPEC_FOLDER_PATH=${DSPEC_FOLDER_PATH}
+      - MODEL_FOLDER_PATH=${MODEL_FOLDER_PATH}
+      - HOHONU_API_AUTH=${HOHONU_API_AUTH} 
+      - HOHONU_API_HOST=${HOHONU_API_HOST}
+    ports: 
+      - "${FRONTEND_HOST_PORT}:8080"
```
matdenkas, line 106: This needs to be added to .env.dist

Dockerfile.frontend
```diff
@@ -0,0 +1,20 @@
...
+RUN npm install
+
+# Copy the rest of the source code
+COPY ./src/Frontend/semaphore-ui/ ./
+
+#Expose the port
+EXPOSE 8080
+
+# Start the Vue.js development server
+CMD ["npm", "run", "serve"]
```
matdenkas, line 20: Do we not need to run build first? This might also need to be added to our CI/CD script, we certainly shouldn't be putting the built frontend on git hub. Ive not used vue so I could be wrong about having to build it.

Dockerfile.frontend
```diff
@@ -0,0 +1,20 @@
...
+WORKDIR /app
+
+# Copy the package.json and package-lock.json
+COPY ./src/Frontend/semaphore-ui/package.json ./package-lock.json ./
+
+# Install dependencies
+RUN npm install
+
+# Copy the rest of the source code
+COPY ./src/Frontend/semaphore-ui/ ./
```
matdenkas, line 14: Putting this here but it doesn't have to do with this line. Under frontend there is .src/Frontend/ChartJson/ ... I feel this would be better under the ./docs folder like how the other dspecs are. No reason for this to block this PR though.

matdenkas (changes requested):
Nice work, this is all very exciting. To much future front-ending 🥂 


Just a few changes about the CI/CD pipeline, not the most important stuff though:
- port missing from env.dist
- We might have to build before running the server
- Careful about exposing tones of envs, no we are not most secure but with things like authentication token, it should always be need to know.

lovelysandlonelys (approved):
Well done!!!! From what I can understand everything looks good to me and page displayed as expected.

Dockerfile.frontend
abeasley1722: Makes sense

Dockerfile.frontend
abeasley1722: We will need to build in the production version of the dockerfile. The serve command handles everything needed for Vue's development server.

src/Frontend/semaphore-ui/public/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="">
```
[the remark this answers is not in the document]
abeasley1722: Good catch. This was automatically made by Vue. I forgot t add en.

Dockerfile.frontend
```diff
@@ -0,0 +1,21 @@
+# Use node image to build and run Vue.js development server
+FROM node:14 AS build
```
snguyen1, line 2: Consider moving to new node version since 14 is being deprecated:
https://endoflife.date/nodejs
Also, some Vue + TypeScript versions only compatible with some certain node versions.

snguyen1 (changes requested):

src/Frontend/semaphore-ui/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "serve": "vue-cli-service serve --host 0.0.0.0",
+    "build": "vue-cli-service build",
+    "lint": "vue-cli-service lint"
+  },
+  "dependencies": {
+    "core-js": "^3.8.3",
+    "vue": "^3.2.13"
```
snguyen1, line 12: may need to update to a newer version of Vue. Depends on what version of node we're going for.

snguyen1 (changes requested):

Dockerfile.frontend
```diff
@@ -0,0 +1,21 @@
...
+
+# Set the working directory
+WORKDIR /app
+
+# Copy the package.json and package-lock.json to the current working directory
+COPY ./src/Frontend/semaphore-ui/package.json ./
+COPY ./package-lock.json ./
```
snguyen1, line 9: Ideally, package.json and package-lock.json should be in the same folder on host. Looks like package-lock.json was copied from the root working directory instead of `.src/Frontend....` like package.json

snguyen1 (changes requested):

PR #482, opened 2024-10-17
latest outputs endpoint

matdenkas:
~~I added a new endpoint called /latest_ouputs its queries outputs based just on a model name and time range, inferring things like model version and lead time based on the latest model version in the db and the latest ouput from that model.~~

~~As a note swaggy ui, what fast api uses for its docs seems to have a bug or something rn, idk I didnt super look into it, but testing through the docs page isn't working. You can use this instead, just make sure you fill it out correctly~~

~~`curl 'http://localhost:8888/semaphore-api/latest_outputs/modelName=ar_inundation_apr_12h/fromDateTime=2024101618/toDateTime=2024101819'  -H 'accept: application/json' -v`~~

### DIFF
--- a/src/API/apiDriver.py
+++ b/src/API/apiDriver.py
@@ -85,7 +85,7 @@ async def get_input(source: str, series: str, location: str, fromDateTime: str, 
 @app.get('/output_latest/')
 async def get_outputs_latest(modelNames: list[str] = Query(None)):
     """
-    Queries outputs for a for given models looking for the last prediction that was made with them.
+    Queries outputs for a given models looking for the last prediction that was made with them.
     Args:
         - `modelNames` (string): The name of the model (e.g. "test AI"), you can repeat this parameter to request multiple models
 
--- a/src/API/apiDriver.py
+++ b/src/API/apiDriver.py
@@ -10,7 +10,7 @@
 # 
 #
 from dotenv import load_dotenv
-from fastapi import FastAPI, HTTPException
+from fastapi import FastAPI, HTTPException, Query
 
 from datetime import datetime, timedelta
 from DataClasses import SeriesDescription, SemaphoreSeriesDescription, TimeDescription
@@ -77,9 +77,53 @@ async def get_input(source: str, series: str, location: str, fromDateTime: str, 
     )
 
     provider = SeriesProvider()
-    responseSeries = provider.request_input(requestDescription, timeDescription)
+    responseSeries = provider.request_input(requestDescription, timeDescription, saveIngestion=False)
 
     return responseSeries
+
+
+@app.get('/output_latest/')
+async def get_outputs_latest(modelNames: list[str] = Query(None)):
+    """
+    Queries outputs for a for given models looking for the last prediction that was made with them.
+    Args:
+        - `modelNames` (string): The name of the model (e.g. "test AI"), you can repeat this parameter to request multiple models
+
+    Returns:
+        The results for each model index by the model name. If no data can be found for that model the value will be null.
+    """ 
+
+    provider = SeriesProvider()
+    results = {}
+    for modelName in modelNames:
+        results[modelName] = provider.request_output('LATEST', model_name= modelName)
+    return results
+
+
+@app.get('/output_time_span/')
+async def get_outputs_time_span(fromDateTime: str, toDateTime: str, modelNames: list[str] = Query(None)):
+    """
+    Queries outputs for a given time range for given models.
+    Args:
+        - `fromDateTime` (string): "YYYYMMDDHH" Date to start at
+        - `toDateTime` (string): "YYYYMMDDHH" Date to end at
+        - `modelNames` (string): The name of the model (e.g. "test AI"), you can repeat this parameter to request multiple models
+
+    Returns:
+        The results for each model index by the model name. If no data can be found for that model for the provided time range the value will be null.
+    """ 
+    try:
+        fromDateTime = datetime.strptime(fromDateTime, '%Y%m%d%H')
+        toDateTime = datetime.strptime(toDateTime, '%Y%m%d%H')
+    except (ValueError, TypeError, OverflowError) as e:
+        raise HTTPException(status_code=404, detail=f'{e}')
+
+    provider = SeriesProvider()
+    results = {}
+    for modelName in modelNames:
+        results[modelName] = provider.request_output('TIME_SPAN', model_name=modelName, from_time=fromDateTime, to_time=toDateTime)
+    return results
+
 
 
 @app.get('/output/modelName={modelName}/modelVersion={modelVersion}/series={series}/location={location}/fromDateTime={fromDateTime}/toDateTime={toDateTime}')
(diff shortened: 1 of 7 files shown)


matdenkas:
Second round of changes is done. There are two sets of changes in this PR. First Sereies Provider request input has a parameter that lets you choose if it should insert into the db or not. Second the API has been updated to have three total endpoints about requesting outputs. There is outputs (which is legacy and untouched) output_latest, and output_time_span. The new endpoints are updated to work with docs and you can fill information there. How I tested was first running pytest but then:

1. Get some data to work with
```docker exec semaphore-core python3 src/semaphoreRunner.py -d ./Inundation/April/ar_inundation_apr_12h.json -p 202410292000
docker exec semaphore-core python3 src/semaphoreRunner.py -d ./Inundation/August/ar_inundation_aug_12h.json -p 202410292000
docker exec semaphore-core python3 src/semaphoreRunner.py -d ./Inundation/December/ar_inundation_dec_12h.json -p 202410292000
docker exec semaphore-core python3 src/semaphoreRunner.py -d ./Inundation/March/ar_inundation_mar_12h.json -p 202410292000```
2. Query it with the different endpoints, the model names and time range is below
Model Names:
```ar_inundation_apr_12h
ar_inundation_aug_12h
ar_inundation_dec_12h
ar_inundation_mar_12h```

Time range:
2024103008
2024103008

You can add more data to the DB, manually change things, try the old endpoint do whatever you want but this is how I tested it.

One thing that was cool is I got tired of defining more and more methods in Series provider that were basically just wrappers for method in the ORM, so I refactored it such that there is just one request output methods, that will take any arguments. It will make sure those arguments get to the right method and throws an exception if something went wrong.

File: src/API/apiDriver.py [hunk 1]
```diff
@@ -76,11 +77,55 @@ async def get_input(source: str, series: str, location: str, fromDateTime: str,
...
-    responseSeries = provider.request_input(requestDescription, timeDescription)
+    responseSeries = provider.request_input(requestDescription, timeDescription, saveIngestion=False)
 
     return responseSeries
 
 
+@app.get('/output_latest/')
+async def get_outputs_latest(modelNames: list[str] = Query(None)):
+    """
+    Queries outputs for a for given models looking for the last prediction that was made with them.
```
abeasley1722, line 88: 'for a for' to just 'for a'

src/SeriesProvider/SeriesProvider.py
```diff
@@ -83,17 +83,45 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
         return self.__data_interpolation(seriesDescription, timeDescription, validated_merged_results)
    
     
-    def request_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription, timeDescription: TimeDescription) -> Series: 
-        ''' Takes a description of an output series and attempts to return it
-            :param seriesDescription: SemaphoreSeriesDescription -A semaphore series description
-            :param timeDescription: TimeDescription - A description about the temporal parts of the output series
-            :return series
+    def request_output(self, method: str, **kwargs) -> Series | None:
```
abeasley1722, line 86: I've never seen **kwargs before. Great find. It seems like something that could be useful for us going forward.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -128,8 +156,12 @@ def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescr
 
         if data_ingestion_results is None: return None, None # ingestion returns None if there was an error
 
-        # If we actually got some data, that data should be new and we need to save it in the database.
-        if(data_ingestion_results.data and data_ingestion_results.description.dataSource != "SEMAPHORE"):
+        
+        # Check if we should be saving this data in the db.
+        hasData = len(data_ingestion_results.data) > 0 # If we actually got some data,
```
abeasley1722, line 161: What if the value is nan? Do we still want to save it?

abeasley1722: image

src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
```diff
@@ -102,6 +102,76 @@ def select_output(self, semaphoreSeriesDescription: SemaphoreSeriesDescription,
         series.data = outputResult
         return series
     
+
+    def select_output(self, model_name: str, from_time: datetime, to_time: datetime) -> Series | None:
```
abeasley1722, line 106: That makes sense. Since it's the same model, all instances will share the same datum and so on, so there's no need to specify anything further.

src/tests/IntegrationTests/test_SeriesProvider.py
```diff
@@ -48,5 +48,5 @@ def test_request_output():
     semaphoreSeriesDescription = SemaphoreSeriesDescription('pineapple', 'strawberry', 'apple', 'pear')
     timeDescription = TimeDescription(datetime(2001, 12, 28), datetime(2001, 12, 28) + timedelta(hours=24))
     series = Series(semaphoreSeriesDescription, True, timeDescription)
-    seriesProvider.request_output(semaphoreSeriesDescription, timeDescription)
+    seriesProvider.request_output('SPECIFIC', semaphoreSeriesDescription=semaphoreSeriesDescription, timeDescription=timeDescription)
```
abeasley1722, line 51: This clearly took a lot of time and effort. Great job as always!

abeasley1722:
output latest works as intended, but only null is returned for the output timespan. I ran the models for 17:00, 18:00, and 19:00.
I tried switching the from time an to time. They produce the same results.

URL I used:
http://[localhost:8888/semaphore-api/output_time_span/?fromDateTime=2024110417&toDateTime=2024110419&modelNames=ar_inundation_apr_12h&modelNames=ar_inundation_aug_12h&modelNames=ar_inundation_dec_12h&modelNames=ar_inundation_mar_12h](http://localhost:8888/semaphore-api/output_time_span/?fromDateTime=2024110417&toDateTime=2024110419&modelNames=ar_inundation_apr_12h&modelNames=ar_inundation_aug_12h&modelNames=ar_inundation_dec_12h&modelNames=ar_inundation_mar_12h)

image

src/SeriesProvider/SeriesProvider.py
matdenkas: Hey Joy! 
So the table you're showing there seems like the output table and the code this is on is about saving inputs. 
For the output we recently did a refactor to make sure we saved NANs, there are a few reasons for this but its mostly about record keeping. 
For the input table I don't think we are actually able to save Nan's afaik the schema doesn't let the data value be null. I don't think an explicit check is a bad thing, though I might argue thats more of the ingestion class responsibility then this check here.

src/API/apiDriver.py
matdenkas: Fixed! Thank you, I really appriciate it!

lovelysandlonelys (approved):
image
Tested end points and they both worked as intended, I have a little bit of time so I'm racing against whenever Matthew decides to merge.

PR #484, opened 2024-11-04
Adding in multi running models

matdenkas:
You can test running multiple models like this:

```docker exec semaphore-core python3 src/semaphoreRunner.py -d ./VirginiaKey/12hr_VirginiaKey_wl_Estrada.json ./VirginiaKey/24hr_VirginiaKey_wl_Estrada.json```

### DIFF
--- a/src/semaphoreRunner.py
+++ b/src/semaphoreRunner.py
@@ -40,8 +40,8 @@ def main():
                     epilog='End Help')
    
     #adding argument types to the parser
-    parser.add_argument("-d", "--dspec", type=str, required=True,
-                       help= "The name of the DSPEC file of the model you wish Semaphore to run. Should be a .json file.")
+    parser.add_argument("-d", "--dspec", nargs='+', type=str, required=True,
+                       help= "The name of the DSPEC file of the model you wish Semaphore to run. Should be a .json file. You can supply multiple DSPEC paths split by spaces. Will run the left most first and the rightmost last.")
     
     parser.add_argument("-p", "--past", type=str, required=False, default=None,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/semaphoreRunner.py [hunk 1]
```diff
@@ -191,10 +188,21 @@ def handle_successful_prediction(model_name: str, execution_time: datetime, resu
             log(inserted_results)
             log(inserted_results.data if inserted_results is not None else '')
     except:
-        log(Semaphore_Exception('ERROR:: An error occurred while trying to interact with series storage from semaphoreRunner'))
+        log(Semaphore_Exception('ERROR:: An error occurred while trying to interact with series storage from semaphoreRunner'))    
 
-    log('Semaphore fin!')
-    exit(0)
+
+def clean_and_check_dspec(dspec_path: str) -> str:
```
lovelysandlonelys, line 194: Nice to add a function for this!

lovelysandlonelys (approved):
Ran correctly and documentation looks good to me! One thing we might consider is also changing the file comment to mention that we can now run multiple models but I don't think it's necessary.

src/semaphoreRunner.py
```diff
@@ -122,6 +119,8 @@ def handle_failed_prediction(
                             input_gatherer: InputGatherer | None, 
                             toss: bool
                             ):
+    """Handles a failed predictions. This function handles semaphore in an unstable error state. It emmits a notification and
```
lovelysandlonelys, line 122: Handles a failed predictions -> Handles failed predictions.

matdenkas: Nice catch, however, we run these in sequence so this function would still only handle one instance. Unless ive misunderstood what you mean.

PR #495, opened 2024-12-19
Cold stunning

matdenkas:
This PR adds a DSPEC for the cold stunning team. Lead times are: `[3,6,12,18,24,30,36,42,48,54,60,66,72,78,84,90,96,102,108,114,120]`
Also added is the H5 for lead time: 96 (renamed in the files to our naming conventions)

The name picked for the cold stunning dspecs is longer than 25 characters so I increased the modelName field in the db to 50.
DB version 2.7 is claimed in this PR. An **invasive** deletion is used for the rollback. It will delete all entries in the model_runs and outputs table for any model with a model name greater than 25 before rolling back the table.

To test:
1.  Build and raise containers
2. `docker exec semaphore-core python3 ./tools/migrate_db.py`
3. `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./ColdStunning/Bird-Island_Water-Temperatue_96hr.json`


lovelysandlonelys (approved):
image
Ran successfully! Didn't find anything wrong, but wondering if we should changed the documentation about the models because it looks like the final input vector predicted air temps isn't the lead time + 6 hours? It's just the lead time into the future?

abeasley1722 (approved):
Ran as expected and code looks good to me!

PR #500, opened 2024-12-20
Adding indexes to 1.0 parser

matdenkas:
I never added indexing to the 1.0 parser it was causing the 1.0 dspecs to fail. This bug patches that, easily testable.

1. Build and run
2. `docker exec semaphore-core python3 src/semaphoreRunner.py -d ThermalRefuge.json`


abeasley1722 (approved):
The prediction was created as intended.

lovelysandlonelys (approved):
image
ran as expected 👍🏻

PR #503, opened 2024-12-29
Added cases that handle actual air and water temperature from NOAATANDC

abeasley1722:
This PR adds cases to NOAATANDC that handle actual air and water temperature.

## How to Test:
1. Remove the statement in test_NOAATANDC that skips the test.
2. Run the tests: docker exec semaphore-core python3 -m pytest -s

I added a print statement to show the data. While it may look messy when testing all tests, the ingestion classes are usually skipped anyway. When testing ingestion classes individually, printing the data is helpful to ensure the data is correct.

## Proof of Completion:
The first screenshots are the data from the test and the last 3 screen shots is data from NOAATANDC. This verifies that the correct data is being retrieved from NOAATANDC. Convert from Celsius to Fahrenheit. ATMP is Air Temperature and WTMP is Water Temperature.
Screenshot 2024-12-28 200813
Screenshot 2024-12-28 200827
Screenshot 2024-12-28 200912
Screenshot 2024-12-28 201107
Screenshot 2024-12-28 201321

### DIFF
--- a/src/DataIngestion/DI_Classes/NOAATANDC.py
+++ b/src/DataIngestion/DI_Classes/NOAATANDC.py
@@ -55,6 +55,10 @@ class NOAATANDC(IDataIngestion):
                 return self.__fetch_WnSpd(seriesDescription, timeDescription)
             case 'dWnDir':
                 return self.__fetch_WnDir(seriesDescription, timeDescription)
+            case 'dAirTmp':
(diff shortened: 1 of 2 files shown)


matdenkas (approved):
Looks good to me, we should use log > print in tests though. Not something that should block this PR

PR #505, opened 2025-01-02
Mapped laguna madre to ndfd_exp

abeasley1722:
The Coldstunning chart requires laguna madre datapoints from NDFD_EXP.


matdenkas (approved):
Migration works!

PR #513, opened 2025-01-21
508 semaphore api returns internal server error when attempting to fetch inputs for flare graph

lovelysandlonelys:
### Issue
When attempting to fetch inputs for the measurements displayed on the Laguna Madre the Semaphore API returns an "Internal Server Error". After further investigation we've found that the API is giving: 2025-01-10 13:52:06 semaphore-api   | ValueError: Out of range float values are not JSON compliant when we attempt to return the inputs from series provider to the API as a JSON object. We've found that this is because there are a small number of nans returned in the data and FastAPI's jsonable_encoder
does not allow nans in the jason object that it returns to the user. 

Our solution was to add a check for nans in the API driver and swap the nans into Nones which FastAPI's jsonable_encoder can parse/remove on it's own. Meaning that the response body that the user gets will exclude the timestamps where a nan was found. 

### Test

1. Bring up the containers. 
2. Check that you can get a response body from the request input method. The below screenshot is the use case we were looking at. 
image
If you're test gets back the response "Internal Service Error" we need to keep working, but if you get values then we should be set for Flare to get measurements.

### DIFF
--- a/src/API/apiDriver.py
+++ b/src/API/apiDriver.py
@@ -81,14 +81,9 @@ async def get_input(source: str, series: str, location: str, fromDateTime: str, 
     responseSeries = provider.request_input(requestDescription, timeDescription, saveIngestion=False)
 
     # Change the nans in the responseSeries to Nones because fast api's jsonable_encoder cannot handle nans
-    print(f"Checking for Nans in the resposne series")
     for value in responseSeries.data:
-        print(f"Value: {value.dataValue}")
-        print(f"Value Type: {type(value.dataValue)}")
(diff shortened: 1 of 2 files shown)


matdenkas (approved):
Nicely done. Thank yall for fixing this. This is a patchwork solution though and kinda bad. 
Datavalue is supposed to be a string. But the parser things its a float. This also means there is a blind cast happening in np.nan() which should be fine. But this just all feels a bit messy. We need to move forwards though.

abeasley1722 (approved):
Works as intended!

PR #518, opened 2025-01-22
Fixed the 2.9 rollback and added a function that deletes rows from mapping

abeasley1722:
I added a function that specifically deletes rows from mapping without deleting rows from other reference tables. It uses both dataLocation and dataSource since there are many locations with the same name, but different sources.

## How to test (Make sure you start at version 2.9):

Step 1: Build the docker containers
         - `docker compose down`
         - `docker compose build`
         - `docker compose up -d`

Step 2: Open pgadmin and check the datasource_datalocation_mapping table. If you are on version 2.9 there will be a lagunamadre - NDFD_EXP mapping. at the bottom.
Screenshot 2025-01-22 133154

Step 3: Change the target version in Semaphore to 2.8 to test roll back from 2.9 to 2.8 

Step 4: Build the docker containers

Step 5: Run this command: `docker exec semaphore-core python3 ./tools/migrate_db.py`

Step 6: Close the datasource_datalocation_mapping table tab that you have open and refresh the database. 
           - Right click the database to refresh
           
Step 7: Check the table again and the mapping should be gone

Screenshot 2025-01-22 132944

### DIFF
--- a/tools/DatabaseMigration/2_9/2_9_DatabaseMigration.py
+++ b/tools/DatabaseMigration/2_9/2_9_DatabaseMigration.py
@@ -6,7 +6,7 @@
 # Version 1.0
 #----------------------------------
 """This is a database migration script that will initialize version
-    2.9 of the database. The change from version 2.9 to 2.9 will add
+    2.9 of the database. The change from version 2.8 to 2.9 will add
     rows to the locations reference table and the location
     mapping reference table for Cold Stunning air and water temperature Measurements. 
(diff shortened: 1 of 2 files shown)


matdenkas (approved):
Nice world solving this fast and big plus for improving the migrator. This is a clean and maintainable solution.

Worked updating and rolling back for me

lovelysandlonelys (approved):
works as expected! thank you joy!!

PR #528, opened 2025-01-28
515 plan scheduler improvement

matdenkas:
Refactoring init_cron.py to schedule models differently. Now we group by interval and offset and order by lead time. We also process this into intermediate files where now cron will trigger a helper script that will process the correct semaphore command and trigger that. This is to overcome the character limit cron files have.

# To test:
Run: `wsl` You have to run this script in linux as it deals with cron
Run: `python3 ./tools/init_cron.py -r ./data/dspec -i ./schedule --max_dspec_per_command 10`
Notice: The new schedule folder with the intermediate files.
Run: `crontab -l` and you should se the correct crontab for the jobs


abeasley1722 (commented):
THis is the resulting cron file initialized. I don't understand how cron will use this with place holder  [insert_helper_script_command]  is still there.
image

abeasley1722:
THis looked like a decent amount of code to write and seems to work how you intended, so well done! Just have one question about the final cron.

abeasley1722 (approved):
Works as intended!

abeasley1722:
> THis is the resulting cron file initialized. I don't understand how cron will use this with place holder [insert_helper_script_command] is still there. image

EDIT: Matthew has informed me that the commands will be added in the next PR.

PR #529, opened 2025-01-30
526 generate helper script to link cron jobs semaphore and intermediate files

matdenkas:
This should finish up the scheduling refactor to create a full pipeline using intermediate files.

# To Test:

1. First: Rebuild Semaphore
2. Run: `wsl` You have to run this script in linux as it deals with cron
3. Run: `python3 ./tools/init_cron.py -r ./data/dspec -i ./schedule --max_dspec_per_command 10`
4. Notice: The new schedule folder with the intermediate files.
5. Run: `crontab -l` and you should se the correct crontab for the jobs
6. Now you can actually run one of those jobs by:
7. Run: `python3 ./tools/group_runner.py -i ./schedule/job_1320_3600_0.json`

* Notice as part of this PR I removed semaphore_runner's assumption that a dspec will be found in ./data/dspec/ Instead a proper absolute or relative path is required by the CLI. The way semaphore runner was doing it kept us from allowing to run dspecs from multiple sources.


abeasley1722 (approved):
Works as intended. Very cool idea and execution of it. The crontab looks so much better!

jeremiahsosa77 (approved):
Functions correctly, and is really cool to see how you did this!

PR #530, opened 2025-01-31
523 prediction requests in ndfd throw xml error when dates are in the past

abeasley1722:
This addition will add a function that validates the time passed to NDFD_EXP.

How to test:
- Build containers
- Use this URL to test that the function returns false if either datetimes are in the past: http://localhost:8888/docs

### DIFF
--- a/src/DataIngestion/DI_Classes/NDFD_EXP.py
+++ b/src/DataIngestion/DI_Classes/NDFD_EXP.py
@@ -60,9 +60,7 @@ class NDFD_EXP(IDataIngestion):
 
         sleep(30)
 
-        validation_result = date_validation(timeDescription)
-        if not validation_result:
-            return
+        date_validation(timeDescription)
 
         # Remove digits 
         processed_series = re.sub('\d', '', seriesDescription.dataSeries)
@@ -445,9 +443,8 @@ def date_validation(timeDescription : TimeDescription) -> bool:
 
     now = datetime.now().replace(minute=0, second=0, microsecond=0)
 
-    if from_datetime < now or to_datetime < now:
-        log("ERROR: Invalid Date Time Provided. API request not granted.")
-        log(f'From Time:{from_datetime} To Time:{to_datetime}')
-        return False
-    
-    return True
+    try:
+        if from_datetime < now or to_datetime < now:
+            raise ValueError(f'Invalid Date Time Provided. Ingestion request cannot execute')
+    except Exception as err:
+        log(f'Unexpected error occured: {err} (could be an API or ingestion model)')
--- a/src/DataIngestion/DI_Classes/NDFD_EXP.py
+++ b/src/DataIngestion/DI_Classes/NDFD_EXP.py
@@ -34,6 +34,7 @@ from DataIngestion.IDataIngestion import IDataIngestion
 from DataClasses import Series, SeriesDescription, Input, TimeDescription
 from SeriesStorage.ISeriesStorage import series_storage_factory
 from utility import log
+from exceptions import Semaphore_Ingestion_Exception
 import re
 import traceback
 import os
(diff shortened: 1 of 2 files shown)


src/DataIngestion/DI_Classes/NDFD_EXP.py
```diff
@@ -430,4 +436,18 @@ def iso8601_to_unixms(timestamp: str) -> int:
...
+    
+def date_validation(timeDescription : TimeDescription) -> bool:
+    """Checks if date time passed is valid"""
+    to_datetime = timeDescription.toDateTime
+    from_datetime = timeDescription.fromDateTime
+
+    now = datetime.now().replace(minute=0, second=0, microsecond=0)
+
+    if from_datetime < now or to_datetime < now:
+        log("ERROR: Invalid Date Time Provided. API request not granted.")
```
matdenkas, line 449: This class isn't necessarily and API call, it could also be ingestion from a model.

matdenkas (changes requested):
The ticket states that this should throw an Error. Maybe not a Semaphore Error but  Data Ingestion Error. Was there a reason that you decided to just log?

I think the log statements should be changed to reflect that Ingestion classes are abstracted from either being part of the API or Ingestion process.

src/DataIngestion/DI_Classes/NDFD_EXP.py
```diff
@@ -430,4 +434,17 @@ def iso8601_to_unixms(timestamp: str) -> int:
...
+        raise ValueError(f"Error converting timestamp to milliseconds: {e}")
+    
+def date_validation(timeDescription : TimeDescription) -> bool:
+    """Checks if date time passed is valid"""
+    to_datetime = timeDescription.toDateTime
+    from_datetime = timeDescription.fromDateTime
+
+    now = datetime.now().replace(minute=0, second=0, microsecond=0)
+
+    try:
```
matdenkas, line 446: This try loop is throwing and catching its own error! So this does nothing.

Also you should be using an internal exception. Look in `./src/exceptions.py`

matdenkas (changes requested):
This works, it logs that there was an invalid date time provided. A few things though:

- The logging is vague, we can say that a time before now time was provided.
- There is no error thrown, its caught in its own loop. I believe a cleaner solution is to throw an error such that the archaic error generated by the parser never occurs. 
- If you were to throw an error it should be one of semaphore's exception classes, see the comment on the try loop.

Reading the ticket back, it says only logging is required, im not sure thats the best solution. But youre technically doing that and we need to get this ticket down so ima say this isn't blocking though id highly prefer things to be implemented with exceptions. Feel free to talk to others on the team or even bring it up in the meeting! This is your call.

lovelysandlonelys (changes requested):
When testing the logs give this error: 
image

I think that this is referring to the fact that in the implementation of Semaphore_Ingestion_Exception in NDFD_EXP sys.exc_info() returns (None, None, None) because no exception is currently being handled. This means exc_tb is None, causing an AttributeError when accessing exc_tb.tb_frame.

If there's an exception during the execution of a block of code in a try/except block, a traceback object is generated, which gives you information about where the error occurred. If you manually raise an exception and don't have an active traceback context, sys.exc_info() will return None for the traceback, leading to this error. In this case, exc_tb will be None because the exception was manually raised and wasn't the result of a preceding exception being caught. 

In order to use the semaphore ingestion exception outside of a try/except block we could Modify the __init__ method of our exception class to handle cases where the traceback (exc_tb) is None. This can be done by checking if exc_tb is None before trying to access its attributes.

image

The above fix would give you this error with some proceeding information that I think is mostly the result of terminating in the middle of an API request: 
image

We could also avoid this by changing the use of the exception to be within a try/except block but I'm not sure if we want to do this since it seems like we went out of our way to avoid that?

lovelysandlonelys (approved):
image

error message is what we wanted and semaphore stops!!

PR #531, opened 2025-02-01
Refactoring logging to be independent of cron nor dependent on a sing…

matdenkas:
# Ticket Explanation
I refactored logging to be non cron dependent and splits logs when multiple models are run.

- I created a singleton class to bridge semaphore_runner needing to be able to set the logging location and me not wanting to refactor every point where we called the log method. The log method will always reference the singleton.
- I set the log base path as an enviroment variable. It will always need to be in a mounted directory as Semaphore is running inside a container and we need the logs outside. I set it in the ./data (./app/data) directory as its already a mounted volume. The env will need to be set on deployment.

# To Test
1. Build containers
2. Run multiple models ` docker exec semaphore-core python3 src/semaphoreRunner.py -d MLP-OP.json ./Inundation/March/ar_inundation_mar_12h.json`
3. Look in your `./data` directory **(Not in the container)** and you should see two different log files formatted correctly in the correct folders.

### DIFF
--- a/src/utility.py
+++ b/src/utility.py
@@ -14,7 +14,48 @@
 from datetime import datetime
 from os.path import isabs
 from os import getcwd
+from os import makedirs
+from os.path import dirname, exists
 
+
+
+class LogLocationDirector(object):
+	"""A singleton class that manages the location of the log file.
+	The logger will always reference this signleton to determine where to write
+	the log file. If the path does not exist it will create it. 
+	"""
+
+	_log_target_path = None
+
+	def __new__(cls):
+		"""A singleton constructor that ensures only one instance of the class"""
+		if not hasattr(cls, 'instance'):
+			cls.instance = super(LogLocationDirector, cls).__new__(cls)
+		return cls.instance
+	
+	@property
+	def log_target_path(self) -> str:
+		return self._log_target_path
+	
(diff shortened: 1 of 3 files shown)


src/utility.py
```diff
@@ -14,25 +14,75 @@
...
 from os.path import isabs
 from os import getcwd
+from os import makedirs
+from os.path import dirname, exists
 
 
 
+class LogLocationDirector(object):
+	"""A singleton class that manages the location of the log file.
+	The logger will always reference this signleton to determine where to write
```
abeasley1722, line 24: singleton*

src/utility.py
```diff
@@ -14,25 +14,75 @@
...
+
+	# Write to log file.
+	log_file = LogLocationDirector().log_target_path
+	if log_file is not None:
+		with open(log_file, 'a') as log_file:
+			log_file.write(f'{msg}\n')
+
 
 def construct_true_path(path: str) -> str:
 	"""A method to determin if a given path is a relative or abolute path. If
```
abeasley1722, line 88: determine*

abeasley1722 (approved):
Works as intended. Great work as always! Just a few spelling mistakes.

PR #533, opened 2025-02-04
Updating deployment script to make it able to rollback

matdenkas:
See ticket for issues.

Updated deployment script to fetch and prune tags from origin, and then specifically checkout the tag that needs to be active.


abeasley1722 (approved):
Essential addition for rollback!

PR #534, opened 2025-02-04
Removing type hints that don't work with the python version on the se…

matdenkas:
The issue with my code was not environment variables, I got myself confused last night. Its because there is an earlier version of python on the server. Type hinting is different between 3.6(server) and 3.11(local).

I solved this by just removing type hinting from the script. I tested it on dev to ensure it would work. Please look it over and then we can re-deploy.


abeasley1722 (approved):
LGTM Great job solving this fast

PR #535, opened 2025-02-10
Changing assumed interval from 6min to 1hour

matdenkas:
This is harder to test.

- Run an API query with not interval requested
- Run some models

Most of the time interval is explicitly requested. CBOCP does request hourly data explicitly so no matter what thats safe.


abeasley1722 (approved):
Small but much needed change. Thanks!

lovelysandlonelys (approved):
was able to run models and request from api with expected results

Issue #555, opened 2025-02-26
Refactor DSPEC


lovelysandlonelys:
* We want to add a field to the order vector that provides a hint that a dependent series is a multivalve and a verification number.
* Refactor The Parser to include this new hint
* Have a test DSPEC just to ensure that parsing method is still backwards compatible.

PR #557, opened 2025-02-26
549 modify input gatherer to format the inputs in an array rather than a single vector

matdenkas:
# What I did
I split inputGatherer into:
   inputvectorBuilder
   dataGatherer

## input vector builder:
    - Generating one or more input vectors required by the specification
    - Casting the data to the correct type
    - Selecting the correct indexing of the data for the input vector
    - Iterate over multi value inputs to generate a batch of vectors

## data gatherer:
    - Request the data from the series provider
    - Post process the data if it is specified
    - Return a dictionary of the data

As a note, I refactored the constructors and these classes expect to be passed DSPEC objects in preparation for our refactors.
I did not write the tests yet. I have started working on them, but I am mildly blocked by #555 to write my tests in a nice way.

# Dependencies 

## @abeasley1722 
image
VectorOrder is an object from the dspec parser. I am expected two feilds
multipliedKeys: A list of keys that we expect their inputs to have multiple values, empty if none exist
amntExpectedVectors: The number of input vectors expected at the end. This can be 1 or None as a default, this is used as a sanity check.

NOTE:: These aren't final, they are just what I came up with and we can def talk about it if you have different ideas.

## @jeremiahsosa77 and @lovelysandlonelys 
I wrote my system expecting that Input.dataValue would either be:
- a data value (Ex. `float: input.dataValue = 1.1`)
- a list of values (Ex. list[float] ` list[float]: input.dataValue = [1.1, 2.2, 3.3]`) 
To put this another way, a single value is never expected to be wrapped in an array.
NOTE:: These aren't final, they are just what I came up with and we can def talk about it if you have different ideas.


lovelysandlonelys (approved):
Tests past and things look like they make sense to me! Neat to use a generator, I haven't seen/worked with one before. 
image
one note about the tests is that it breaks some of the other tests that inport the old inputGatherer class.

PR #558, opened 2025-02-27
553 refactor input and output objects

jeremiahsosa77:
# **What We Did**

### **The DataClasses file**
Has been modified to handle both single float values and lists of float values for input and output. Now, if a single float is provided, it is treated as such. If a list of floats is provided, it will be processed as a list. So, even if a single value is wrapped in an array, it will no longer cause issues and should be able to handle both cases appropriately.
***

### **The test_DataClasses file**
Has been modified to test single float values and list of float values for input and output but was also updated to test other classes within the DataClasses file.

***
# **How To Test**
 docker-compose up -d --build
 docker exec semaphore-core python -m pytest

***

## **Notes**
When running test, one test does fail but, test is from test_inputGatherer test and is most likely not from anything added by us

Sav and I are pretty sure that this is all that was needed but please advise us if we missed something!

### DIFF
--- a/src/DataClasses.py
+++ b/src/DataClasses.py
@@ -23,7 +23,7 @@ class Input():
         :param latitude: str = None
 
     """
-    def __init__(self, dataValue: float | list[float], dataUnit: str, timeVerified: datetime, timeGenerated: datetime,  longitude: str = None, latitude: str = None) -> None:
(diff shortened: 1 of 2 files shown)


matdenkas (approved):
LGTM! Fucking grade A job for writing all those tests!!! Slowly we will crawl semaphore closer to being a respectable body of code!!

abeasley1722 (approved):
LGTM! Good Job!

PR #560, opened 2025-03-03
552 refactor dspecparser inputgatherer seriesprovider and modelwrapper to follow orchestrator

matdenkas:
This is the refactor implementing an Orchestrator and returning Semaphore runner to just being a CLI. This is not testable yet as its still dependent on the modelWrapper refactor. However once that done that should fix dataGatherer and we should be ready for integration testing.


src/orchestrator.py
```diff
@@ -0,0 +1,213 @@
+# -*- coding: utf-8 -*-
+#semaphoreRunner.py
```
lovelysandlonelys, line 2: incorrect file name

lovelysandlonelys (approved):
I think this looks good to me, mostly just moving things around right? Nice job! I had one documentation error but doesn't necessarily need to block merging to the feature branch.

PR #565, opened 2025-03-11
563 refactor dspec again to have numberensablemembers in the input vector

abeasley1722:
Changed the variable name as asked  and made sure that only one vector can be an ensemble.
## How to Test:

- Build the containers
- run this command to run tests:  `docker exec semaphore-core python3 -m pytest -s`
- Ensure the dspec parser test passes

### DIFF
--- a/src/ModelExecution/dspecParser.py
+++ b/src/ModelExecution/dspecParser.py
@@ -111,7 +111,7 @@ class dspec_sub_Parser_1_0:
             types = []
             indexes = []
             multipliedKeys = []
-            amntExpectedVectors = []
+            ensembleMemberCount = []
             
             for idx, inputJson in enumerate(inputsJson):
                 dseries = DependentSeries()
@@ -132,8 +132,8 @@ class dspec_sub_Parser_1_0:
                 types.append(inputJson["type"])
                 keys.append(str(idx))
                 indexes.append((None, None))
-                multipliedKeys.append(None)
-                amntExpectedVectors.append(None)
+                multipliedKeys.append([])
+                ensembleMemberCount.append(None)
                 
 
                 dependentSeriesList.append(dseries)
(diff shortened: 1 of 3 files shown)


src/ModelExecution/dspecParser.py
```diff
@@ -244,21 +252,48 @@ def __parse_vector_order(self):
...
+                    
+            
 
+            
         vectorOrder = VectorOrder()
         vectorOrder.keys = keys
         vectorOrder.dTypes = dTypes
         vectorOrder.indexes = indexes
+        vectorOrder.multipliedKeys = multipliedKeys
+        vectorOrder.ensembleMemberCount = ensembleMemberCount
```
matdenkas, line 294: This would be type list[int | None] As we have the requirement that there can be only one, shouldn't this just be the count?

src/ModelExecution/dspecParser.py
```diff
@@ -244,21 +252,48 @@ def __parse_vector_order(self):
...
+                    raise ValueError("Error: More than one multipliedKey has been detected!")
+                found_multipliedKeys = True
+            multipliedKeys.append(multipliedKey) 
+
+            ensembleMember: int | None = dict.get("ensembleMemberCount", None)  # Default to None
+            if "ensembleMemberCount" in dict:
+                if found_ensembleMemberCount:
+                    raise ValueError("Error: More than one ensembleMemberCount has been detected!")
+                found_ensembleMemberCount = True
+            ensembleMemberCount.append(ensembleMember)
```
matdenkas, line 284: Instead of having two other variables to track if we have found this or not, we can check the size of lists or if a variable isn't none anymore

matdenkas (changes requested):
I think this solution needs a little more workshopping, could be wrong feel free to overwrite

PR #579, opened 2025-03-15
566 data frame refactor

matdenkas:


data/dspec/TestModels/test_multiVector.json
```diff
@@ -0,0 +1,164 @@
+{
+    "dspecVersion": "2.0",
+    "modelName": "test_multiVector",
```
abeasley1722, line 3: Great work as always, it's clear you put a lot of time and effort into this! The implementation is clear making it easy to follow. 

A few assertions failed from the test, but the models worked as intended


Screenshot 2025-03-28 111352

matdenkas: Fixed, I had to clean up the parsing a bit but I think its in a good place now if you wanna take another look

lovelysandlonelys (approved):
Tests passed: 
image
Ran a few models and they went smoothly as well.

Issue #581, opened 2025-03-18
Update Queries for Ensemble DB refactor.


matdenkas:
Queries in the DB need to be updated:

- [x] Insert Input needs to detect an ensemble input, adding groups to the value before inserting from their implicit order.
- [x] Insert Outputs needs to detect an ensemble output, adding groups to the value before inserting from their implicit order.
- [x] Select Outputs needs to return grouped data correctly. ***
- [x] Select Input needs to return grouped data correctly. ***

*** Under a normal request all multiple rows should be returned over a timespan and ensemble. Thus they should be ordered by timespan each time span would have multiple rows that need to be ordered into one input in the order of their group number. The default behavior of normal data needs to be preserved.

PR #586, opened 2025-03-20
3.0 ensembleMemberID column migration

matdenkas:
This PR updates the DB in a migration script to have the ensembleMemberID in both the inputs and outputs table. It also updates the AK for both tables to include this column.

# To test:

### Test forwards:
```bash
docker compose up --build -d
```
Using PGAdmin you can view the properties of the tables to check their conditions, types, and name

## Test backwards:
Alter `.\tools\DatabaseMigration\target_version.json` to the previous version of 2.12.
```bash
docker compose up --build -d
```

You can also go a step further and run a model to have some data in the DB before you migrate. You will see that all the rows get updated with ensembleMemberID being null (default behavior)

### DIFF
--- a/tools/DatabaseMigration/3_0/3_0_DatabaseMigration.py
+++ b/tools/DatabaseMigration/3_0/3_0_DatabaseMigration.py
@@ -1,0 +1,145 @@
+# -*- coding: utf-8 -*-
+#3_0_DatabaseMigration.py
+#----------------------------------
+# Created By: Matthew Kastl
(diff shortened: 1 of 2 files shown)


matdenkas:
This has been merged to #581 and should not be merged

abeasley1722 (approved):
Works as intended!

***Update**

The tests worked as intended, but when I run models this error shows up:
image

jeremiahsosa77 (approved):
Looks good, updates and rolls back as intended!

PR #594, opened 2025-03-25
544 create the weather company ingestion class

matdenkas:
The weather company ingestion class. As a few notes, 1. Not sure about the time range. TWC includes an inclusive range and I just couldn't get my head around if thats right or not. I think this is right but it will be easier to test with a DSPEC. Second I used the pro-type feature, we will have to ask tissot if that is right. Finally its names TWC because data sources have a 10 character limit.
### **You need the API key to test this, you can text Savannah or I for the key**
You can test the class logic yourself with something like below:
```python3
from DataIngestion.DI_Classes.TWC import TWC
from DataClasses import SeriesDescription, TimeDescription
from datetime import datetime, timedelta
from dotenv import load_dotenv
load_dotenv()


# Create a TimeDescription object
time_description = TimeDescription(
    fromDateTime=datetime.utcnow(),
    toDateTime=datetime.utcnow() + timedelta(hours=3),
    interval=timedelta(hours=1)
)

# Create a SeriesDescription object
series_description = SeriesDescription(
    dataSource="NOAATANDC",
    dataSeries="temperature",
    dataLocation="lagunamadre",
    dataDatum="sea_level",
    dataIntegrityDescription=None,
    verificationOverride=None
)

TWC = TWC()
result = TWC.ingest_series(
    seriesDescription=series_description, 
    timeDescription=time_description
)

print(result)
print(result.dataFrame.to_string())
```

 You can also run the tests here with: 
`docker exec semaphore-core python3 -m pytest src/tests/UnitTests/test_TWC.py `


src/tests/UnitTests/test_TWC.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
+#test_TWC.py
+#-------------------------------
+# Created By: Matthew Kastl
+# version 1.0
+#----------------------------------
+""" This provides unit tests for the The weather company ingestion class
```
abeasley1722, line 7: "the The" to the

abeasley1722 (approved):
Worked as intended! Great work!

PR #596, opened 2025-03-25
docker deploy

jeremiahsosa77:
## OverView
Updating Semaphores deployment script to match version 28 of docker which doesn't use the dash in 'docker-compose'
using 'docker compose' now instead of 'docker-compose'

## Testing
docker compose build
docker compose up -d

LMK if i did anything wrong :)

### DIFF
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -7,7 +7,7 @@
 
 
 # Lower active containers
-docker-compose down
+docker compose down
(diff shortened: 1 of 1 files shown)


matdenkas (approved):
This is good, we need to push this up soon!

lovelysandlonelys (approved):
👍🏻

PR #615, opened 2025-04-09
Ensemble interpolation

matdenkas:
Creating a data Integrity class that can preform linear interpolation for ensemble data.

To test:
`docker exec semaphore-core python3 -m pytest src/tests/UnitTests/test_EnsemblePandasInterpolation.py`

### DIFF
--- a/src/DataIntegrity/DataIntegrityClasses/EnsemblePandasInterpolation.py
+++ b/src/DataIntegrity/DataIntegrityClasses/EnsemblePandasInterpolation.py
@@ -1,0 +1,102 @@
+# -*- coding: utf-8 -*-
+# EnsemblePandasInterpolation.py
+#----------------------------------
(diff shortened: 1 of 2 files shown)


src/DataIntegrity/DataIntegrityClasses/EnsemblePandasInterpolation.py
```diff
@@ -0,0 +1,102 @@
+# -*- coding: utf-8 -*-
+# EnsemblePandasInterpolation.py
+#----------------------------------
+# Created By: Op Team
```
lovelysandlonelys, line 4: one thing about matthew is that he will do a slightly different header on every file ( i like this style tho so don't change it)

lovelysandlonelys (approved):
image
tests passed and everything looked good to me

PR #627, opened 2025-04-28
The api can now handle an array of data from the database

abeasley1722:
## How to test
- Build the containers
- Go to this browser: http://localhost:8888/docs#/default/get_outputs_latest_output_latest__get
- Test output_latest with 'MRE_Bird-Island_Water-Temperature_3hr'


lovelysandlonelys (approved):
output api works for both sigle result model and mutiresult model and code looks good to me

PR #628, opened 2025-04-28
API TWC Error

lovelysandlonelys:
A fix for the error where when requesting TWC data we would get a `ValueError: Out of range float values are not JSON compliant`.  This fix changes it so that we make the dictionary, replace NaN's with None and then encode the response because unlike NaNs, None's can be handled by the encoder. 

To Test: 
- Build and run containers. 
- Visit local API: [Input Request](http://localhost:8888/docs#/default/get_input_input_source__source__series__series__location__location__fromDateTime__fromDateTime__toDateTime__toDateTime__get)
- Request TPW data similar to below screenshot but altered to use timestamps int he future: 
image


abeasley1722 (approved):
LGTM!

PR #643, opened 2025-05-13
Im a big dummy dumb dumb

matdenkas:
The missing entries are caused by lists of dspecs being run in a loop. This loop was nested into a try block, meaning if any model failed the try block would catch it and control flow never returns back to the loop. Ergo the fix was to move the loop out of the try block.

To test you can mess up a dspec in one of the job so it will crash, and notice that all the models still get run


lovelysandlonelys (approved):
logic is sound

abeasley1722 (approved):
LGTM!

PR #661, opened 2025-07-08
Removed $ from lat/long in url

CJQuintero61:
To run:

1. docker compose down
2. docker compose up --build -d
3. go to http://localhost:8888/docs#/default/get_input_input_source__source__series__series__location__location__fromDateTime__fromDateTime__toDateTime__toDateTime__get
4. fill in required fields
5. check the output in dataValue: [...]. Celsius values should be around high 20's and low 30's

I recommend copy and pasting a dataValue block into an AI to find the median and convert it to fahrenheit to see that the temperatures are much higher now. 

image
image


matdenkas (approved):
Forgor to submit this, my B! Everything looks good and works fine

PR #670, opened 2025-07-21
added blank variables for account username and uid

CJQuintero61:
Added blank fields to the env.dist file for the service account.


CJQuintero61:
@snguyen1 is this first task really that simple? Just want to make sure before I move onto changing the next sub task.

Issue #681, opened 2025-08-26
Deprication Warnings Due to Version Upgrades


matdenkas:

ccftissot:
consolidate with #936 ?

Issue #680, opened 2025-08-26
Fix version issues for all libraries and all containers


ccftissot:
There is one issue currently that breaks semaphore-core (see below), but we should take this opportunity to:

1.  update all of libraries to the most recent version that still works with our code
2.  change requriments.txt to use fix versions instead of minimum versions

### Current issue
1. we use the tensorflow 2.14 image to create the semaphore-core container
2. we constrain tensorflow to a _minimum_ version of 2.14.
3. Today tensorflow is at version 2.20,  So, when our containers build, we use the 2.14 images but then install tf 2.20 and the 2 are not compatible (version 2.19 is still compatible).  It looks like 2.20 was compiled (from C++) with different settings that clash with the 2.14 libraries.

### What we need to do
1. update the core container to the most recent tensorflow image that will run our code w/o changes
2. update the reqruirements.txt file to use that fixed version 
3. check for the latest version of all other libraries we use (pandas, numpy, etc.) with which our current code will still run and update the requirements.txt file to use these fixed versions
4. Do the same thing for the DB and API containers (ensure we are using the most recent version that still works with our code and use fixed versions of these libraries)

matdenkas:
Numpy 2.x is not out of the box compatible with the psycopg 2.x driver. I upgraded Semaphore to the 3.x driver. NOTE:: this caused a env change b/c the db connection strings need to reflect that we are using the 3.x driver now

ccftissot:
> First order requirements:
> 
> ```
> fastapi== 0.116.1
> uvicorn==0.35.0
> lxml== 6.0.1
> httpx==0.28.1
> numpy==2.3.2
> pytest==8.4.1
> python-dotenv==1.1.1
> requests==2.32.5
> SQLAlchemy==2.0.43
> psycopg==3.2.9
> psycopg-binary==3.2.9
> pandas==2.3.2
> noaa-coops==0.4.0
> discord-webhook==1.4.1
> # tensorflow==2.20.0 -- Provided by base image
> ```

interesting that we don't have tf in the first order requirements?  i see it in the full list though

ccftissot:
> Issue: A second order dependency "protobuf" is sending a user error about 20 times (once for each tf sub library):
> 
> ```
[...]

PR #682, opened 2025-08-26
Version Update

matdenkas:
# Why:
The TF versions in DEV an PROD were mismatched. Along with fixing that this was a good time to update ALL of our dependencies. 

# What:
 - I updated the base image for semaphore-core to the latest version of Tensorflow
 - I altered semaphore-api to be based on just a python base as it shouldn't need TF
 - I updated each of the dependencies to their latest version

# Anything change?
Yes! Most things were updated and our pytorch driver got changed from psycopg2 -> psycopg3. This matters because it changes the DB connection string, something stored in your ENVs. Thus **YOU HAVE TO UPDATE YOUR .ENV** and dev and prod will need theirs updated too. 
     Simply change: `postgresql+psycopg2://` to `postgresql+psycopg://` (You should have to do this in more than one place if you have @lovelysandlonelys DB user environment variables which you should)

# How did you test?
1) I ran tests (they all passed)
2) I ran a few models (trying to run models that I knew would call different code)
3) I made sure I could pull data from the API
4) I ran a lot of models using 
```
import pandas as pd
import os
import subprocess

TIME =  '202508262000'
directory = './data/dspec'

# Walk through the directory
for root, dirs, files in os.walk(directory):
    for file in files:
        if file.endswith('.json') and file != 'comment.json':
            file_path = os.path.join(root, file)
            file_path = file_path.replace("\\", "/")
            print(file_path)

            # Run the command using subprocess
            command = [
                "docker", "exec", "semaphore-core", 
                "python3", "src/semaphoreRunner.py", 
                "-d", file_path, 
                "-p", TIME
            ]
            print(" ".join(command))            
            try:
                result = subprocess.run(command, check=True, text=True, capture_output=True, encoding='utf-8')
                print("Command output:", result.stdout)
            except subprocess.CalledProcessError as e:
                print("Error occurred:", e.stderr)
```

# How should I test?
-  Check the containers build correctly `docker compose up --build -d`
- Please run the tests `docker exec semaphore-core python3 -m pytest`
- Please run some models
- Please check you can pull data from the API (We need some tests for this)

Keep an eye out for `User Wanrings:`/`Deprication Warnings:` I have detailed the ones ive found in a [ticket ](https://github.com/conrad-blucher-institute/semaphore/issues/681)but there could be some I missed!

For more info please see the original [ticket](https://github.com/conrad-blucher-institute/semaphore/issues/680).

@snguyen1 I added you as a reviewer because the base python image which is from the official source had some vulnerability warnings flash up in my IDE. I just hoped you would take a quick look and make sure its safe, thanks!


matdenkas:
> image @matdenkas

This is the testing result, scroll up and you will see why the test failed, those are the errors I need

abeasley1722 (commented):
Same tests failed: 
image

PR #708, opened 2025-09-16
Adding DataValidation as its own decoupled module

matdenkas:
# What:
Moving data validation out of series provider gives us the opportunity to refactor it and decouple it and properly test it. That is what this work contains.

# What I did:
- Created a new IDataValidation interface and factory to abstract the process of data validation
- Created a DateRangeValidation class to be the default validation via the time description (refactored to be cleaner)
- Crated a OverrideValidation class to be used by aggregate use cases
- Added tests for the two new classes
- I did NOT remove the methods or tests from series provider to avoid merge conflicts.

# How to test?
3. See all tests pass (Except maybe the float error 🔪)


CJQuintero61 (approved):
good

PR #709, opened 2025-09-19
691 twc sets generated time

CJQuintero61:
### How it works
TWC does not give an explicit generated time, so we will use initTime extracted from the metadata. Only a single initTime is given, so all rows use the same initTime. This then gets converted using .utcfromtimestamp( ) to get our generatedTime.

### What was changed
- TWC ingestion class now sets generated time as the returned initTime from the response
- updated unit tests to include an initTime key/value pair 
- added an assertion loop to ensure all rows have the same timeGenerated set

### How to test
- `docker compose down`
- `docker exec semaphore-core python3 -m pytest -v` 
- All tests should pass (except the 2 model_runner assertion tests that fail due to the trailing decimal if you get them).


lovelysandlonelys (approved):
tests passed and nice documentation on change!
image
(just the floating point one which is.. showing up for me now...)

PR #712, opened 2025-09-23
707 refactor data ingestion classes to remove reindexing

lovelysandlonelys:
All ingestion classes were checked for re-indexing and re-indexing was removed if needed. I also made some fixes to depreciated time methods that I figure since we're touching a lot of files with the refactor we should fix as we go? If we don't want to do that let me know and I can swap them back. 

To test I think just run the tests since we won't expect models to run correctly with the ingestion classes not doing indexing.


abeasley1722 (approved):
Works as intended!

PR #714, opened 2025-09-24
702 refactor series storage to remove reindexing in selection queries

### LINKED ISSUE #702: Refactor Series Storage to remove reindexing in selection queries
## [This is part of the planned refactor ](https://github.com/conrad-blucher-institute/semaphore/issues/688)

We no longer are reindexing or removing interval before input gatherer. The selection queries in the ORM do remove out of interval rows and that code needs to be removed sending all rows found downstream.

lovelysandlonelys:
Really Simple to remove re-indexing, just check that tests still pass while we're in the transition phase and data gather hasn't been refactored yet to handle getting all the available data.  Double check that I didn't miss anything I should have removed please. Will close #702.

### DIFF
--- a/src/DataIngestion/DI_Classes/LIGHTHOUSE.py
+++ b/src/DataIngestion/DI_Classes/LIGHTHOUSE.py
@@ -3,7 +3,7 @@
 #----------------------------------
 # Created By: Matthew Kastl
 # Created Date: 11/3/2023
-# version 1.0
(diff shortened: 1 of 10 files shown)


abeasley1722:
tz aware error.  ALthough, tests pass. We need to discuss whether we change everything to tz aware or keep it naive, but that should be another ticket.
image

PR #715, opened 2025-09-26
Added acquired time staleness

abeasley1722:
## Why? 
We need to ensure that we only ingest data if data is out of the freshness window. I set that to an hour, for now. It works by finding the earliest acquired time for that set of data and if it is older than an hour from now then we initialize data ingestion.

# How I tested
- Added these comments: 
```
def __determine_staleness(self, validated_db_results: Series, reference_time: timedelta) -> bool:
        """
        Returns True (fresh) if the OLDEST acquiredTime in the series is within the allowed
        freshness window of the reference time; otherwise returns False (stale).

        Expected attributes:
        - validated_db_results.timeDescription.referenceTime: datetime-like
        - validated_db_results.dataFrame: pandas DataFrame with column 'acquiredTime'
        """

        freshness_window = timedelta(hours=1)

        df = validated_db_results.dataFrame
        if df is None or len(df) == 0:
            return False
        
        print(f'[DF] {df.to_string()}')

        # Coerce acquiredTime to datetimes and get the OLDEST (minimum) timestamp
        acq = pd.to_datetime(df["acquiredTime"], errors="coerce")
        oldest_acq = acq.min(skipna=True)
        print(f'[Oldest Acquired time] {oldest_acq}')
        if pd.isna(oldest_acq):
            return False
        
        print(f'[reference_time ] {reference_time}')

        # Normalize tz-awareness (use timezone-naive comparison)
        reference_time = pd.to_datetime(reference_time)
        if isinstance(reference_time, pd.Timestamp) and reference_time.tz is not None:
            reference_time = reference_time.tz_localize(None)
        if isinstance(oldest_acq, pd.Timestamp) and oldest_acq.tz is not None:
            oldest_acq = oldest_acq.tz_localize(None)
            
        print(f'[reference_time 2] {reference_time}')
        print(f'[Oldest Acquired time 2] {oldest_acq}')

        # Age = how far the OLDEST acquired sample is from the reference
        age = reference_time - oldest_acq
        print(f'[AGE] {age}')

        # If the oldest sample is no older than the freshness window, we're good
        return age <= freshness_window
```

## How you can test:

- docker compose up --build -d
- docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_12hr.json
-  docker exec semaphore-core python3 -m pytest -s


CJQuintero61 (changes requested):
I don't think Series Provider can use the .isComplete var anymore since Matthew removed that variable from Series objects in another PR. LMK if I'm in the wrong.

abeasley1722:
Replaced by another branch

PR #716, opened 2025-09-27
706 refactor series storage to add a query just for aquired time

abeasley1722:
## Why? 
We need to ensure that we only ingest data if data is out of the staleness window. It works by finding the latest (furthest from reference time) acquired time for that set of data and if it is older than the window then we initialize data ingestion.

## Changes
- Removed .iscomplete from remaining files
- Updated series provider  test
- Created a function that determines stalenes
- Created a test for the determine staleness function
- Created a function that replaces is_complete to determine if we recieved enough data from the database. If not, we initiate data ingestion.
- Removed interval formatting from select input
- Updated splice input to account for acquired time

Note:
We may need to remove interval formatting fully from series provider in the future.  That messes with multiple other things, so I viewed that as out of scope for this task.

# How I tested
- Added these comments(Only paste the comments if need be): 
```
def __determine_staleness(self, validated_db_results: Series, reference_time: timedelta) -> bool:
        """
        Returns True (fresh) if the OLDEST acquiredTime in the series is within the allowed
        freshness window of the reference time; otherwise returns False (stale).

        Expected attributes:
        - validated_db_results.timeDescription.referenceTime: datetime-like
        - validated_db_results.dataFrame: pandas DataFrame with column 'acquiredTime'
        """

        freshness_window = timedelta(hours=1)

        df = validated_db_results.dataFrame
        if df is None or len(df) == 0:
            return False
        
        print(f'[DF] {df.to_string()}')

        # Coerce acquiredTime to datetimes and get the OLDEST (minimum) timestamp
        acq = pd.to_datetime(df["acquiredTime"], errors="coerce")
        oldest_acq = acq.min(skipna=True)
        print(f'[Oldest Acquired time] {oldest_acq}')
        if pd.isna(oldest_acq):
            return False
        
        print(f'[reference_time ] {reference_time}')

        # Normalize tz-awareness (use timezone-naive comparison)
        reference_time = pd.to_datetime(reference_time)
        if isinstance(reference_time, pd.Timestamp) and reference_time.tz is not None:
            reference_time = reference_time.tz_localize(None)
        if isinstance(oldest_acq, pd.Timestamp) and oldest_acq.tz is not None:
            oldest_acq = oldest_acq.tz_localize(None)
            
        print(f'[reference_time 2] {reference_time}')
        print(f'[Oldest Acquired time 2] {oldest_acq}')

        # Age = how far the OLDEST acquired sample is from the reference
        age = reference_time - oldest_acq
        print(f'[AGE] {age}')

        # If the oldest sample is no older than the freshness window, we're good
        return age <= freshness_window
```

## How you can test:

- docker compose up --build -d
- docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_12hr.json
-  docker exec semaphore-core python3 -m pytest -s


CJQuintero61 (changes requested):
Image

Code looks good, but im getting an error for test_API. ```FAILED src/tests/IntegrationTests/test_API.py::test_getInput[NOAATANDC-dXWnCmp000D-packChan-2023090411-2023090511-dXWnCmp]```

CJQuintero61:
Just tested unit tests on one of my branches, and the test_API one is failing there, so it's not specific to your changes. Please disregard it.

PR #725, opened 2025-10-06
706 insure api is serializing data objects correctly

### LINKED ISSUE #717: Change datetime methods to timezone aware
We're currently using a deprecated method for creating UTC time stamps. To use a non deprecated method the datetime library uses "timezone aware" datatype so we need to carefully comb through the files to ensure everything is comparable and setting times the way they were before updating methods.

lovelysandlonelys:
# About
To close #717 
Even though Series.isComplete no longer exists, every API response will still include "isComplete": true in its JSON output so that downstream services expecting isComplete will continue to work unchanged. The change is limited to serialization only, which means our ingestion, processing, and API endpoints stay untouched.

# To Test
I'll be for real guys!! Not really a way to test this because it depends so much on the other refactors (that I... have not reviewed yet). This is something that we'll have to check as we merge things I think. But in theory (I think) this should ensure that isComplete is marked as true so that downstream things that use semaphore data (Flare) don't have to worry about the fact that the isComplete is no longer a part of the series object.

### DIFF
--- a/src/API/apiDriver.py
+++ b/src/API/apiDriver.py
@@ -200,6 +200,7 @@ def serialize_input_series(series: Series) -> dict[any]:
     param: series - Series The input series to serialize.
     return: A dictionary representation of the input series.
 
+    NOTE:: Ensures `isComplete` always exists and is set to True for backward compatibility.
     NOTE:: Expected input series serialization
     {
         "description": {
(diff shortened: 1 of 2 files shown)


CJQuintero61 (approved):
Good

PR #734, opened 2025-10-16
Updating transform modle to run at 31 minuets past the hour

matdenkas:
# What?
We are altering the schedule for the magnolia models to run at 31 minuets past the hour so the premodels have a chance to finish before we try and run them.

# To test
You can try running the models but nothing really was changed.


abeasley1722 (approved):
LGTM. Does what it intends to.

CJQuintero61 (approved):

PR #741, opened 2025-10-21
Fixing Attribute error

matdenkas:
You can run:
`docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Inundation/July/ar_inundation_jul_12h.json`
and
`docker exec semaphore-core python3 -m pytest`


CJQuintero61:
image

same issue as Sav, and agreed that this bug is not related to this ticket. Tests pass.

PR #742, opened 2025-10-21
Fixing bug in text and refixing NOAATANDC

matdenkas:
So what was happening here is that NOAA gives back timezone unaware datetimes. In the test code however, very cool-y the NOAA code is mocked out. Unfortunately the mocked version uses timezone AWARE datetimes. So the fix here is to change the code in NOAATANDC to expect time zone unaware datetimes and change the tests to use timezone unaware datetimes. We do convert to time zone aware later down the line before ingesting in semaphore.

To test run:
`docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Surge/North_Jetty/nj_mlp_surge_12h_pred.json`
and the tests:
`docker exec semaphore-core python3 -m pytest`

### DIFF
--- a/src/DataIngestion/DI_Classes/NOAATANDC.py
+++ b/src/DataIngestion/DI_Classes/NOAATANDC.py
@@ -167,8 +167,8 @@ class NOAATANDC(IDataIngestion):
             log(f'NOAA COOPS invalid request error: {e}')
             return None
 
-        if isSinglePoint:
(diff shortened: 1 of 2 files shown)


lovelysandlonelys (approved):
fix makes sense and ran on my machine

CJQuintero61 (approved):
good

PR #743, opened 2025-10-22
bug fix to handle more regex versions

lovelysandlonelys:
run:  docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/VirginiaKey/24hr_VirginiaKey_wl_Estrada.json


CJQuintero61 (approved):
good, model runs and tests pass

PR #767, opened 2025-11-12
765 fix db has data in time range

CJQuintero61:
# Problem
We want to figure out if we should ingest new data based on the max verified time for the time range. 

# Solution
Series Storage now only returns the row with the greatest verified time for the given time range. 
The check for whether or not we should ingest data is now done in Series Provider where we will ingest data if the verified time is strictly < the toDateTime AND the (referenceTime - acquiredTime) is strictly > threshold. 

# Changes
* Series Storage returns the entire row with the max verified time for a given range
* Series Provider checks the returned row to decide if ingestion should be called
* new unit test file for the new logic
* removed old tests from test_unit_series_storage.py 

# How to test
* `docker compose up --build -d`
* `docker exec semaphore-core python3 -m pytest` 
* ensure all unit tests pass
* `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_120hr.json`
* `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Inundation/June/ar_inundation_jun_24h.json`

Whichever models you choose to run, make sure to run them twice because likely ingestion will occur the first time, then the DB query without ingestion will happen the second time. Ensure that models run correctly when ingesting and when they do not have to ingest.

# Ticket description
Restating the problem in more details (old description kept for historical purposes):
Objective:

The purpose of this query is to help us decide whether we should call the ingestion class to try and get more data. Not fresher data; we don't care how fresh the data is. We also don't care if we have all of the data; again that is not the role of this part of business logic. We only care about whether we could get more data if we called the ingestion class.

Calling a data source will only yield to more data points within our time range if we fall within 2 possible scenarios:

1. The data source has acquired more data dated after the last data we got from it. For measurements, this means that it has acquired more recent measurements since we last called it. For predictions, this means that it generated predictions with a greater lead time compared to the last time we called it.
2. The data source has revised or acquired past values that were not available at the time we last called it, not because they are new values but because there is some process on the data source side, some QAQC, that ran and updated values for previous verified times that did not have any.

For the purpose of the logic we are trying to implement, we can ignore scenario 2 above as we don't know of any data sources that actually do this. YAGNI. If we ever run into a data source that does that, we can always revisit this logic.
Recommended logic:

Given that we are only trying to support scenario 1, we are only trying to see if we could get greater-dated verified times from our data source that we have in our DB. To do this, we can do the following:

1. get the input row with the max verified time in our time description range (so between our from and to date)
2. if the max verified time is the same as our todate, we are good, calling the ingestion class is not necessary
3. If the max verified time is less than our todate, then we need to decide whether calling the data source would lead to additional verified times. For this we can use the acquired time of our max verified time input.
4. if the acquired time is less than a given threshold as specified in our dspec, or a default if not specified in the dspec, then do not call the data source again.
5. If the acquired time is greater than our threshold then call the data source again.

Recommended Implementation:

1. Modify the query (and the method name and signature) to return the max verified time input from the DB for the given series description and time range. Return all columns for that input.
2. In the series provider, implement the logic describe above. Note that we do not have to modify the dspec for now, just pick a default and implement it. We can discuss what that default should be based on potential NDBC issue with wave statistics. In the future, we can add an attribute to our dspec to specify the threshold. Alternatively, we could use the existing interval attribute as our threshold.

Original issue description:

* The query needs to be updated to not request a range of data just the verified time we want to make sure is there.
* We really should rename this method to make it clearer
* We need to fix the comment that talks about requesting. We should just say there is no data in the database

### DIFF
--- a/src/SeriesProvider/SeriesProvider.py
+++ b/src/SeriesProvider/SeriesProvider.py
@@ -4,7 +4,7 @@
 # Created By: Matthew Kastl
 # Updated By: Anointiyae Beasley
 # Created Date: 4/30/2023
-# Updated Date: 10/02/2025
+# Updated Date: 11/27/2025
 # version 3.0
 #----------------------------------
 """This class is the start point for interacting with the data section of semaphore. All data requests should go through here.
@@ -58,6 +58,9 @@ class SeriesProvider():
         
         reference_time = datetime.now(timezone.utc)
 
+        # assume we have not ingested data yet
+        already_ingested_data = False
+
         # If the data source is from the semaphore ingestion class, we ignore the default behavior and always request new data.
         if seriesDescription.dataSource.upper() == 'SEMAPHORE':
             return self.__data_ingestion_query(seriesDescription, timeDescription)
@@ -65,13 +68,20 @@ class SeriesProvider():
         # We request new data if:
         #   - The data in the db is stale.
         #   - We can get more verified times for the requested range.
+
+        # always call the db freshness check
         db_is_fresh = self.seriesStorage.db_has_freshly_acquired_data(seriesDescription, timeDescription, reference_time)
-
-
-        should_ingest_for_verified_time = self.__check_verified_time_for_ingestion(seriesDescription, timeDescription, reference_time)
-
-        if not db_is_fresh or should_ingest_for_verified_time:
+        
+        if not db_is_fresh:
             self.__data_ingestion_query(seriesDescription, timeDescription)
+            already_ingested_data = True
+
+        # if we haven't ingested for staleness, check verified time ingestion
+        if not already_ingested_data:
+            should_ingest_for_verified_time = self.__check_verified_time_for_ingestion(seriesDescription, timeDescription, reference_time)
+
+            if should_ingest_for_verified_time:
+                self.__data_ingestion_query(seriesDescription, timeDescription)
 
         return self.__data_base_query(seriesDescription, timeDescription)
     


src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
```diff
@@ -471,28 +483,36 @@ def db_has_data_in_time_range(self, seriesDescription: SeriesDescription, timeDe
...
+        # only bind dataDatum if it's not None
         if seriesDescription.dataDatum is not None:
             bind_params['dataDatum'] = seriesDescription.dataDatum
+        
         query_stmt = query_stmt.bindparams(**bind_params)
         
         tupleishResult = self.__dbSelection(query_stmt).fetchall()
-        if not tupleishResult: # Data is not yet in the DB so we need to request it
+
+        print(f'Fetched tuplishResult for db_has_data_in_time_range: {tupleishResult}')
```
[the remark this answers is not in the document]
CJQuintero61: removed print statement

src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
```diff
@@ -451,18 +451,30 @@ def db_has_freshly_acquired_data(self, seriesDescription: SeriesDescription, tim
...
         """
-        Returns true if the database has data up to the toTime specified in the TimeDescription. This means 
-        that the database isn't missing new data.
-
-        Expected attributes:
-        :param seriesDescription: SeriesDescription - A series description object
-        :param timeDescription: TimeDescription - A hydrated time description object
+        This function checks if the DB has data for the requested toTime.
+        It will query the db and check for the greatest verified time that is <= the requested toTime.
+        This is because the greatest verified time for the requested toTime may not be exaclty the requested toTime.
```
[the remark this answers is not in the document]
CJQuintero61: fixed typo

src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
```diff
@@ -451,18 +451,30 @@ def db_has_freshly_acquired_data(self, seriesDescription: SeriesDescription, tim
...
+
+        This is done by 
+            1. selecting the verified times
+            2. filtering by the verified times that are less than or equal to the requested toTime
+            2. ordering them descending (greatest first)
+            3. limiting to 1 result to get the greatest verified time
+            4. checking if the returned verified time is >= the requested toTime
```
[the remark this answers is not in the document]
CJQuintero61: fixed step numbering

src/SeriesStorage/ISeriesStorage.py
```diff
@@ -61,7 +61,7 @@ def db_has_freshly_acquired_data(self, seriesDescription: SeriesDescription, tim
         raise NotImplementedError()
     
     @abstractmethod
-    def db_has_data_in_time_range(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> bool:
+    def db_has_requested_to_time(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> bool:
```
abeasley1722: Good change

src/SeriesProvider/SeriesProvider.py
```diff
@@ -19,7 +19,8 @@
 from exceptions import Semaphore_Ingestion_Exception, Semaphore_Exception
 
 from utility import log
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timedelta
```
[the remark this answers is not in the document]
CJQuintero61: removed timedelta import

src/SeriesProvider/SeriesProvider.py
```diff
@@ -150,4 +154,44 @@ def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescr
...
+        acquired_time = row[2].replace(tzinfo=timezone.utc)
+        toDateTime = timeDescription.toDateTime.replace(tzinfo=timezone.utc)
+
+        # the threshold is set to the interval
+        threshold = timeDescription.interval
+
+        difference = reference_time - acquired_time
+
+        return (verified_time < toDateTime and difference > threshold)
+
```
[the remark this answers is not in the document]
CJQuintero61: added if/else statement for defaulting to 1 hour if interval is None

src/SeriesProvider/SeriesProvider.py
```diff
@@ -60,10 +60,13 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
...
-        #   - The db does not have data newly made data. (Made by data source)
+        #   - We can get more verified times for the requested range.
         db_is_fresh = self.seriesStorage.db_has_freshly_acquired_data(seriesDescription, timeDescription, reference_time)
-        db_has_new = self.seriesStorage.db_has_data_in_time_range(seriesDescription, timeDescription)
-        if not db_is_fresh or not db_has_new:
+
+        max_verified_time_row = self.seriesStorage.fetch_row_with_max_verified_time_in_range(seriesDescription, timeDescription)
+        should_ingest_for_verified_time = self.__check_verified_time_for_ingestion(seriesDescription, timeDescription, reference_time, max_verified_time_row)
+
+        if not db_is_fresh or should_ingest_for_verified_time:
```
ccftissot: we discussed splitting the calls and the checks to avoid making unnecessary calls to the DB.  so something like:

```

db_is_fresh = ...
if not db_is_fresh:
   max_verified_time_row  = ...
   should_ingest_for_verified_time = ...
  if should_ingest_for_verified_time:
      # calling ingest class

# call select_inputs query
```

src/SeriesProvider/SeriesProvider.py
```diff
@@ -150,4 +153,44 @@ def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescr
...
+
+        :param seriesDescription: SeriesDescription - The description for a series
+        :param timeDescription: TimeDescription - The time description for a series
+        :param reference_time: datetime - The reference time of the model execution
+        :param row: tuple - The row with the max verified time in the requested range
+
+        :returns bool 
+
+        True (should ingest) if:
+        - No row exists (database is empty)
```
ccftissot: hmm that doesn't mean the DB is empty necessarily :-).  we just don't have inputs for the provided series description

src/SeriesProvider/SeriesProvider.py
```diff
@@ -150,4 +153,44 @@ def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescr
...
+        if not row:
+            return True
+        
+        # extract times from the row and add timezone info
+        verified_time = row[3].replace(tzinfo=timezone.utc)
+        acquired_time = row[2].replace(tzinfo=timezone.utc)
+        toDateTime = timeDescription.toDateTime.replace(tzinfo=timezone.utc)
+
+        # the threshold is set to the interval if it exists, otherwise default to 1 hour
+        threshold = timeDescription.interval if timeDescription.interval is not None else timedelta(hours=1)
```
ccftissot: is interval an existing attribute in the dspec?  the one that specifies the interval within our time range?  If so, is that really what we want to use ?  it seems to me that would never be none because it is required, no?

src/SeriesProvider/SeriesProvider.py
```diff
@@ -150,4 +153,44 @@ def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescr
             if(inserted_series is None or len(inserted_series.dataFrame) == 0): # A sanity check that the data is actually getting inserted!
                 log('WARNING:: A data insertion was triggered but no data was actually inserted!')
         
-        return data_ingestion_results
\ No newline at end of file
+        return data_ingestion_results
+    
+    def __check_verified_time_for_ingestion(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription, reference_time: datetime, row: tuple) -> bool:
```
ccftissot: another alternative would be to call the query from this function instead of from the main function.  This way we wouldn't have to pass the row as an argument.  Also, logically, this function needs that information to perform its logic so it should be responsible for getting it.

src/SeriesStorage/ISeriesStorage.py
```diff
@@ -61,7 +61,7 @@ def db_has_freshly_acquired_data(self, seriesDescription: SeriesDescription, tim
         raise NotImplementedError()
     
     @abstractmethod
-    def db_has_data_in_time_range(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> bool:
+    def fetch_row_with_max_verified_time_in_range(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> tuple | None:
```
ccftissot: much better name.  well done!

ccftissot (changes requested):
@CJQuintero61 .  i have a few review comments.  let me know if you want to discuss.  I did not build or test anything.  sorry I am a bit short on time and wanted to get this review done asap.  thanks,

src/SeriesProvider/SeriesProvider.py
CJQuintero61: updated comment to
 "        True (should ingest) if:
        - No rows exists for the provided series and time description"

src/SeriesProvider/SeriesProvider.py
```diff
@@ -150,4 +153,44 @@ def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescr
...
+        """
+        if not row:
+            return True
+        
+        # extract times from the row and add timezone info
+        verified_time = row[3].replace(tzinfo=timezone.utc)
+        acquired_time = row[2].replace(tzinfo=timezone.utc)
+        toDateTime = timeDescription.toDateTime.replace(tzinfo=timezone.utc)
+
+        # the threshold is set to the interval if it exists, otherwise default to 1 hour
```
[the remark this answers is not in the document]
CJQuintero61: class constants now set. the default acquire threshold is set to 1 hour and the default staleness threshold is set to 7 hours

src/SeriesProvider/SeriesProvider.py
CJQuintero61: added the row fetching to the function

lovelysandlonelys (approved):
models ran and unit tests passed, but there are some merge conflicts?

CJQuintero61:
Closing this PR since it is already merged into #794 and that branch will be merged instead.

PR #769, opened 2025-11-13
Updated the freshness function

abeasley1722:
# Objective:
- Retrieve the most recent generatedTime for each verifiedTime.
- Determine the oldest value among those most-recent generatedTime values.
- If the age of this oldest recent generatedTime exceeds the allowed staleness threshold, the data is considered not fresh.

# How did I test?
- Ran the query in pgadmin to see if it gave expected results.
- Also ran additional querys showing all of the data, so that I can ensure the eldest was chosen.
- Added a print statement in the freshness function that showed what exactly was being selected by the query and confirmed it was what was expected.

 # How to test:
- ` docker exec semaphore-core python -m pytest -s  src/tests/UnitTests/test_unit_sqlAlchemy.py`

You can also use this statement to see if its correct, but its a lot simpler to use the test since theres less data to sort through and compare.

```
WITH most_recent_per_verification AS (
SELECT DISTINCT
FIRST_VALUE("generatedTime") OVER (
PARTITION BY "verifiedTime"
ORDER BY "generatedTime" DESC
) as most_recent_gen_time
FROM inputs
WHERE "dataLocation" = 'PortOConnor'
AND "dataSource" = 'LIGHTHOUSE'
AND "dataSeries" = 'pHarm'
AND latitude = '28.4460300'
AND longitude = '-96.3960700'
AND "verifiedTime" BETWEEN '2025-11-11 15:00:00' AND '2025-11-13 02:00:00'
)
SELECT MIN(most_recent_gen_time) as oldest_recent_generated_time
FROM most_recent_per_verification;
```

### DIFF
--- a/src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
+++ b/src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
@@ -408,9 +408,11 @@ class SQLAlchemyORM_Postgres(ISeriesStorage):
         Data is considered fresh if it was acquired within the window of [reference time, staleness offset]. The staleness offset
         is configured for this request in the TimeDescription object. Staleness is a measure with acquired time not verified time.
 
-        Query Assumption:
-        - The data in the "WHERE" section is all we need to query the correct data.
-        - We only care about the worst-case staleness across all verified times in that window.
+        Data Assumptions (in the inputs table):
+        - Every verifiedTime in [from_dt, to_dt] has at least one row in the database.
+        - It’s acceptable to treat the ensemble as fresh if any member is fresh.
+        - If the worst-case generatedTime is fresh, you assume all verifiedTimes are fresh enough.
+        - Freshness is evaluated per verifiedTime, not per ensemble member.
         
            
         Query Summary: 


ccftissot:
Have we tested with MRE models? The current query does not account for ensemble IDs:  is that OK?  If we believe so, let's add why in the documentation requested by CJ.

src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
```diff
@@ -414,40 +414,44 @@ def db_has_freshly_acquired_data(self, seriesDescription: SeriesDescription, tim
...
+        if not tupleishResult: #Data is not present in the DB
             return False
-
-        acquiredTime = pd.to_datetime(tupleishResult[0][0]).tz_localize(timezone.utc)
-        age: timedelta = referenceTime - acquiredTime
+        
+        oldestGeneratedTime = pd.to_datetime(tupleishResult[0][0]).tz_localize(timezone.utc)
+        age: timedelta = referenceTime - oldestGeneratedTime
         stalenessOffset = timeDescription.stalenessOffset
         is_fresh = age <= (stalenessOffset if stalenessOffset is not None else timedelta(hours=7)) # Default staleness offset is 7 hours if not specified
```
ccftissot: Consider moving this logic to the calling component (in our case SeriesProvider) to keep our DB access layer clean of business logic.  Make this method return the freshness date of the data and the series provider decide whether it is fresh enough.

ccftissot (changes requested):

ccftissot:
For posterity and because I spent way too much time getting this queries to work :-p, here's a query to get some sample input data from ensemble members that we can use to test that we are getting the data we expect:

SELECT 
    "verifiedTime",
    "generatedTime",
    "dataValue",
    "acquiredTime",
	"ensembleMemberID",
    row_num
FROM (
    SELECT 
        *,
        ROW_NUMBER() OVER (
            PARTITION BY "verifiedTime" 
            ORDER BY "generatedTime" DESC
        ) as row_num
    FROM inputs AS i
        WHERE "dataLocation" = 'SBirdIsland' 
          AND "dataSource" = 'TWC'
          AND "dataSeries" = 'pAirTemp'
          AND "verifiedTime" BETWEEN '2025-11-26 04:00:00' AND '2025-11-26 06:00:00' 
		  AND "ensembleMemberID" BETWEEN 1 and 4
) ranked
WHERE "ensembleMemberID" BETWEEN 1 and 4
ORDER BY "verifiedTime" DESC,  "generatedTime" DESC, "ensembleMemberID";

Issue #761, opened 2025-11-04
Profiling semaphore-core


matdenkas:
The purpose of this task is to profile semaphore-core to explore our execution speeds and identify any bottlenecks

matdenkas:
# Profiling a python profiler
Python by default comes with a profiler called [cProfile](https://docs.python.org/3/library/profile.html). This makes it pretty trivial to profile a program. We want to profile our programs within there containers but simply appending the profiler to the call does this. Example: `docker exec semaphore-core python3 -m cProfile -o prof.prof src/semaphoreRunner.py -d ./data/dspec/Magnolia/12/magnolia_12.json -t` command will profile semaphore running the Magnolia 12 model and dump it into a file called prof.prof. The prof.prof file is likely just a zip but its not readable. You can use a library called pStat to read it or I used a visualizer program called [snakeviz](https://jiffyclub.github.io/snakeviz/) which I found to be much easier.

snguyen1:
Asked Claude about the __SpliceInput method:

<img width="818" height="628" alt="Image" src="https://github.com/user-attachments/assets/5a593c57-eecf-48ed-8d62-f9a36affeb65" />

matdenkas:
# Semaphore Profiling 2 Electric Boogaloo

## Profiling

@CJQuintero61 and I worked on profiling after the query changes. 

Run time is down from 145.137s to 21.495s 🎊 WOOOOOOOO!

Are largest cumulative function (at 16 seconds) was data_gatherer which spent time evenly split between four parts
<img width="1377" height="458" alt="Image" src="https://github.com/user-attachments/assets/00efd2ec-2b7b-4f4e-906a-b8fc2aee6caf" />

Lets dive into what data_gatherer spent time on. Like I said before there are four main parts:

| Function | Purpose | Cumulative time |
| -------- | -------- | -------- |
| __data_base_query   | select inputs| 5.44s |
| __data_ingestion_query | Data ingestion fetch | 4.38s  |
| db_has_freshly_acquired_data | Data check 1| 3.59s |
| db_has_data_in_time_range | Data check 2| 2.64 |

Notice each of these are I/O operations. 

<img width="1364" height="162" alt="Image" src="https://github.com/user-attachments/assets/dc4be3ee-5dba-41dd-84c9-0df2bbab236b" />

Most further analysis didn't reveal much but we wanted to check db time vs __splice_inputs:

[...]

PR #803, opened 2025-12-03
Adding indexes migration and new query

matdenkas:
## What?
Adding in indexes to the inputs table.
Changing the select inputs query.
Removing the manually placed index on DEV

@ccftissot i am pinging you directly because I changed your query a bit. I believe it should be fine but I wanted to make sure with you.
``` sql
        WITH latest_per_group AS (
            SELECT DISTINCT ON ("verifiedTime", "ensembleMemberID")
                i."id",
                i."generatedTime",
                i."acquiredTime",
                i."verifiedTime",
                i."dataValue",
                i."isActual",
                i."dataUnit",
                i."dataSource",
                i."dataLocation",
                i."dataSeries",
                i."dataDatum",
                i."latitude",
                i."longitude",
                i."ensembleMemberID"
            FROM inputs AS i
            WHERE i."dataSource"   = :dataSource
                AND i."dataLocation" = :dataLocation
                AND i."dataSeries"   = :dataSeries
                AND {'i."dataDatum" = :dataDatum' if seriesDescription.dataDatum is not None else 'i."dataDatum" IS NULL'}
                AND i."verifiedTime" BETWEEN :from_dt AND :to_dt
            ORDER BY
                i."verifiedTime",
                i."ensembleMemberID",
                i."generatedTime" DESC
        )
        SELECT
            "id",
            MIN("generatedTime") OVER () AS "generatedTime",
            "acquiredTime",
            "verifiedTime",
            "dataValue",
            "isActual",
            "dataUnit",
            "dataSource",
            "dataLocation",
            "dataSeries",
            "dataDatum",
            "latitude",
            "longitude",
            "ensembleMemberID"
        FROM latest_per_group
        ORDER BY
            "verifiedTime",
            "ensembleMemberID";
```

The specifics is you had MIN("generatedTime") OVER () AS "minGeneratedTime" which was being returned via the query. I moved it to be in place of the actual generated time to keep the number of columns consistent with what the code expects. I am assuming this does the same thing as the MIN OVER command is still being run. I am just selecting the generated time only one instead of twice.

## To test:
1) build
2) Run: `docker exec semaphore-core python3 tools/migrate_db.py`
3) Check indexes w/ pgadmin.
4) Change the DB target to 3.3 and redo step 1 and 2. Check that the indexes are gon. This validates the index migration.
5) run `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/MRE_Bird-Island_Water-Temperature_120hr.json -v` 
6) If the inputs query works this model should succeeded.
7) If you want to validate the index is removed on dev you can. Use PGAdmin to connect and check the index table

### DIFF
--- a/src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
+++ b/src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
@@ -415,11 +415,9 @@ class SQLAlchemyORM_Postgres(ISeriesStorage):
         ids = [row[0] for row in result]
         return resultSeries, ids
 
-    def db_has_freshly_acquired_data(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription, referenceTime: datetime) -> bool:
+    def fetch_oldest_generated_time(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> datetime | None:
         """
-        Returns true if the database has fresh data for all data described in the request.
-        Data is considered fresh if it was acquired within the window of [reference time, staleness offset]. The staleness offset
-        is configured for this request in the TimeDescription object. Staleness is a measure with acquired time not verified time.
+        Returns the oldest generated time within a time window.
 
         Data Assumptions (in the inputs table):
         - Every verifiedTime in [from_dt, to_dt] has at least one row in the database.
@@ -438,8 +436,6 @@ class SQLAlchemyORM_Postgres(ISeriesStorage):
         Expected attributes:
         :param seriesDescription: SeriesDescription - A series description object
         :param timeDescription: TimeDescription - A hydrated time description object
-        :param referenceTime: datetime - The time data is being requested. Usually, it is now.
-
         """
 
         query_stmt = text(f"""
@@ -491,22 +487,26 @@ class SQLAlchemyORM_Postgres(ISeriesStorage):
   
         return is_fresh
        
-
-    def db_has_data_in_time_range(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> bool:
+    def fetch_row_with_max_verified_time_in_range(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> tuple | None:
         """
-        Returns true if the database has data up to the toTime specified in the TimeDescription. This means 
-        that the database isn't missing new data.
+        This function returns the row with the max verified time in the requested range
 
-        Expected attributes:
-        :param seriesDescription: SeriesDescription - A series description object
-        :param timeDescription: TimeDescription - A hydrated time description object
+        params:
+            seriesDescription: SeriesDescription - A series description object
+            timeDescription: TimeDescription - A hydrated time description object
+
+        Returns:
+            tuple | None - The row with the max verified time in the requested range or None if no data is found
+
+        The returned tuple will have the order of:
+        (id, generatedTime, acquiredTime, verifiedTime, dataValue, isActual, dataUnit, dataSource, dataLocation,
+        dataSeries, dataDatum, latitude, longitude, ensembleMemberID)
         """
-        # region toTime Inclusion Check
 
-        # A query to get the latest verified time in the DB for this series in the requested time range
+        # this query gets the row with the max verified time in the requested range
         query_stmt = text(f"""
         SELECT  
-        i."verifiedTime"
+        i.*
         FROM inputs AS i
         WHERE i."dataSource"   = :dataSource
         AND i."dataLocation" = :dataLocation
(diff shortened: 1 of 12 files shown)


ccftissot (changes requested):
i don't have the time to look a it carefully but need to block this from merging when i can.  i think we got confused between 2 queries.  Please take a look at my comment on the original ticket.  I'll get the review done asap, but stopping it from moving forward for now.

src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
```diff
@@ -62,61 +62,49 @@ def select_input(self, seriesDescription: SeriesDescription, timeDescription : T
...
-        "longitude",
-        "ensembleMemberID"
-           
-        FROM
-            RankedData
-        WHERE
-            rn = 1 -- Grabs the latest genrated time from each ensemble member id and verified time
+        SELECT
+            "id",
+            MIN("generatedTime") OVER () AS "generatedTime",
```
ccftissot: it looks like i created a bit of confusion, sorry about that.  I think you were looking at the DISTINCT ON query in ticket #761 but that query was for the freshness query.  That query requires a CTE  that is pretty much the query we need for selecting inputs and that CTE is then used to extract the *oldest* generated time from all the groups.  When we select the inputs, we want to select the top (most recent) generated time for each group.  That is what the CTE does.  So , for that select_inputs method we can just use the select distinct on internal query; we don't need the CTE and we dont' need the outer select.  So it should look something like this:
SELECT DISTINCT ON ("verifiedTime", "ensembleMemberID")
                i."id",
                i."generatedTime",
                i."acquiredTime",
                i."verifiedTime",
                i."dataValue",
                i."isActual",
                i."dataUnit",
                i."dataSource",
                i."dataLocation",
                i."dataSeries",
                i."dataDatum",
                i."latitude",
                i."longitude",
                i."ensembleMemberID"
            FROM inputs AS i
            WHERE i."dataSource"   = :dataSource
                AND i."dataLocation" = :dataLocation
                AND i."dataSeries"   = :dataSeries
                AND {'i."dataDatum" = :dataDatum' if seriesDescription.dataDatum is not None else 'i."dataDatum" IS NULL'}
                AND i."verifiedTime" BETWEEN :from_dt AND :to_dt
            ORDER BY
                i."verifiedTime",
                i."ensembleMemberID",
                i."generatedTime" DESC

tools/DatabaseMigration/3_4/3_4_DatabaseMigration.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def update(self, databaseEngine: Engine) -> bool:
+        """This function updates the database to version 3.4.
+
+           :param databaseEngine: Engine - the engine of the database we are connecting to (semaphore)
+           :return: bool indicating successful update
+        """
+
+
+        stmt_create_idx_inputs_ordering = text('CREATE INDEX idx_inputs_ordering ON inputs ("verifiedTime", "ensembleMemberID", "generatedTime" DESC);')
```
ccftissot: should we consider a bit more defensive SQL?  like dropping that index if it already exists or only adding it if it doesn't?

tools/DatabaseMigration/target_version.json
```diff
@@ -1,4 +1,4 @@
 {
-    "Target Version" : "3.3",
-    "Description" : "Add NDFD_JSON as a new data source"
+    "Target Version" : "3.4",
+    "Description" : "Add indexes to inputs table for performance improvements."
```
ccftissot: we are only adding one index, no?

ccftissot (dismissed):
note that the freshness query also needs to be updated to use a similar pattern as the new select_input query.  it should look something like this:

WITH latest_per_verification AS (
    SELECT DISTINCT ON (i."verifiedTime")
        i."generatedTime"
    FROM inputs AS i
    WHERE
        i."dataSource" = :dataSource
        AND i."dataLocation" = :dataLocation
        AND i."dataSeries" = :dataSeries
        AND (
            (:dataDatum IS NULL AND i."dataDatum" IS NULL)
            OR i."dataDatum" = :dataDatum
        )
        AND i."verifiedTime" BETWEEN :from_dt AND :to_dt
    ORDER BY i."verifiedTime", i."generatedTime" DESC
)
SELECT MIN("generatedTime") AS oldest_recent_generated_time
FROM latest_per_verification;

we can create another ticket for this, but as long as we have the current version of the freshness query in our code, the DB layer will still be fairly slow or at least significantly slower when we run the models

tools/DatabaseMigration/target_version.json
matdenkas: Fixed

tools/DatabaseMigration/3_4/3_4_DatabaseMigration.py
matdenkas: Added

src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
matdenkas: Both queries are updated and tested. Please look them over. I was confused because I thought the minGenerated time OVER () statement was required to prune out results we didnt want returned. I now realize that the DISTINCT ON keyword will select only a single result from each group. Which is super cool!

PR #807, opened 2025-12-05
Reference time is being set to now in series provider

abeasley1722:
## The problem
In series provider reference time is being set to now.  This works for real-time predictions and measurements, but fails for historical data requests.

## How I decided to solve it
I kept reference time as now, since comparing to now is the best way to know how long the data has been in the database. However, instead of treating all requests the same, we adjust the freshness window based on how far back the request is looking. I believe this is the best route as it limits the further manipulation of other files, while effectively ensuring the freshness of data in the past, present, and future. 

For real-time predictions/measurements:

- fromTime ≈ now → lookback = 0

- max_allowed_age = stalenessOffset (same behavior as before)

For historical requests:

- fromTime is farther in the past → lookback expands the freshness window

- This prevents valid historical data from being marked stale

For future predictions:

- fromTime ≥ now → lookback = 0

- Freshness behaves like the original logic

## Potential Questions
- Why are you using the verified from time? Because fromTime represents the earliest verified time the caller needs.
By anchoring staleness grace to fromTime, we ensure enough freshness window for the historical range requested. 

## How to test
- `docker exec semaphore-core python -m pytest -s`
- docker exec semaphore-core python3 src/semaphoreRunner.py -d data/dspec/ColdStunning/Bird-Island_Water-Temperature_6hr.json

### DIFF
--- a/src/SeriesProvider/SeriesProvider.py
+++ b/src/SeriesProvider/SeriesProvider.py
@@ -139,7 +139,7 @@ class SeriesProvider():
         """
         
         oldest_generated_time = self.seriesStorage.fetch_oldest_generated_time(seriesDescription, timeDescription)
-        
+
         stalenessOffset = timeDescription.stalenessOffset
         if stalenessOffset is None:
             stalenessOffset = timedelta(hours=7) #7 because this accounts for at least 2 data updates for all sources
@@ -152,6 +152,10 @@ class SeriesProvider():
         
         # How far back is the earliest verified time?
         lookback = referenceTime - min_v   # e.g. 24h if min_v = now - 24h
+
+        if lookback < stalenessOffset:
+            lookback = timedelta(0) #Set to 0 so that predictions aren't penalized
+
         
         max_allowed_age = lookback + stalenessOffset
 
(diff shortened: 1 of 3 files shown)


src/SeriesProvider/SeriesProvider.py
```diff
@@ -140,12 +140,28 @@ def db_has_freshly_acquired_data(self, seriesDescription: SeriesDescription, tim
         
         oldest_generated_time = self.seriesStorage.fetch_oldest_generated_time(seriesDescription, timeDescription)
 
+        stalenessOffset = timeDescription.stalenessOffset
+        if stalenessOffset is None:
+            stalenessOffset = timedelta(hours=7) #7 because this accounts for at least 2 data updates for all sources
+
```
matdenkas: If stalenessOffset is None then we shouldnt be making this check. This is because its a measurement and we only care if the measurement is in the DB or not. Which is handled by the next check. The defaulting to seven hours should be handled in the DSPEC parsing step, not by this code here.

src/SeriesProvider/SeriesProvider.py
```diff
@@ -140,12 +140,28 @@ def db_has_freshly_acquired_data(self, seriesDescription: SeriesDescription, tim
...
             return False
         
-        age: timedelta = referenceTime - oldest_generated_time
-        stalenessOffset = timeDescription.stalenessOffset
-        is_fresh = age <= (stalenessOffset if stalenessOffset is not None else timedelta(hours=7)) # Default staleness offset is 7 hours if not specified
+        # Earliest verified time
+        min_v = timeDescription.fromDateTime
+        
+        # How far back is the earliest verified time?
+        lookback = referenceTime - min_v   # e.g. 24h if min_v = now - 24h
```
matdenkas: Clean Code: min_v? I assume you mean min_varified_Time but dont abbreviate it

matdenkas: If you are setting reference time to now, that is **NOT REFERENCE TIME**. Thats what made me start this ticket. Reference time is a defined term for semaphore which is specifically != to now. If the algorithm you are writing needs now, then call it now, calling it reference time will cause confusion down the line.

matdenkas (changes requested):
This is all very confusing and we are totally in the weeds. We can take some time to talk about this next meeting but I dont think this is correct. The reason I created the ticket is because there is a variable being called reference time that is being set to now. Now != reference time, so it looks like a bug. I think you can use Now or reference_time but you have to call the var correctly that or it will be confusing.

I am rejecting because there are some errors and the code could be cleaner in my opinion.

CJQuintero61 (changes requested):
I think the implemented logic is pretty confusing and if we need reference time here, why not just pass the actual reference time?

PR #811, opened 2025-12-12
fixed a bug with output endpoints returning duplicate series data

CJQuintero61:
### Issue
ouput endpoints on the api append 2 copies of the same data, one without TZ info and one with it.

### Solution
I moved the inner for loop to the top of the outer for loop to keep the json safe types then just changed append variable to the result of the inner for loop

### To test
3. `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Inundation/June/ar_inundation_jun_24h.json`
4. `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_120hr.json`

ensure models and pytests pass, then
5. go to http://localhost:8888/semaphore-api/docs#/
6. Run any of the output queries, I ran ar_inundation_jun_24h with output latest.

Ensure the series data block only has 1 element and that the generated time does NOT have tz info

### Expected output 
image

### DIFF
--- a/src/API/apiDriver.py
+++ b/src/API/apiDriver.py
@@ -330,13 +330,6 @@ def serialize_output_series(series: Series) -> dict[any]:
     # Serialize the dataFrame by our own rules
     serialized_data = []
     for _, row in series.dataFrame.iterrows():
-        serialized_data.append({
-            "dataValue":      jsonable_encoder(row['dataValue']),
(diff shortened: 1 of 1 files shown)


abeasley1722 (approved):
LGTM

PR #812, opened 2025-12-14
upgraded the freshness function

abeasley1722:
## Objective
The reference time is incorrectly set to now for Semaphore and the freshness query does not handle a None stalenessOffset.
## Changes
- Set reference time to now for api requests
- Passed the correct referenceTime used in Semaphore to the freshness query
- Handled stalenessOffset default in the apiDriver.
- Handled none inside the freshness function
- Removed business logic
- Updated comments for the freshness function
- Ensured the freshness rule works with historical runs
- Updated the test case
- Changed names to match the surrounding name formats
## How to Test
- `docker exec semaphore-core python3 tools/migrate_db.py`
- `docker exec semaphore-core python -m pytest -s src/tests/UnitTests/test_unit_sqlAlchemy.py`
- `docker exec semaphore-core python3 src/semaphoreRunner.py -d data/dspec/ColdStunning/Bird-Island_Water-Temperature_6hr.json`
- See if data was returned: `curl -X GET "http://localhost:8888/semaphore-api/input/source=LIGHTHOUSE/series=dAirTmp/location=SouthBirdIsland/fromDateTime=2025121205/toDateTime=2025121212?interval=3600"`

### DIFF
--- a/src/SeriesProvider/SeriesProvider.py
+++ b/src/SeriesProvider/SeriesProvider.py
@@ -53,6 +53,7 @@ class SeriesProvider():
         """This method attempts to return a series matching a series description and a time description.
             :param seriesDescription - A description of the wanted series.
             :param timeDescription - A description of the temporal information of the wanted series. 
+            :param refrenceTime - The time of execution 
             :returns series - The series containing as much data as could be found.
         """
         log(f'\nInit input request from \t{seriesDescription}\t{timeDescription}')


src/SeriesProvider/SeriesProvider.py
```diff
@@ -49,15 +49,13 @@ def save_output_series(self, series: Series) -> Series:
         return returningSeries
           
     
-    def request_input(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription) -> Series:
+    def request_input(self, seriesDescription: SeriesDescription, timeDescription: TimeDescription, referenceTime: datetime) -> Series:
         """This method attempts to return a series matching a series description and a time description.
             :param seriesDescription - A description of the wanted series.
             :param timeDescription - A description of the temporal information of the wanted series.
```
lovelysandlonelys: We'll have to add refrenceTime to the list of parameters in the function description.

PR #813, opened 2025-12-15
Add monitoring capabilities to the postgres

abeasley1722:
## Objective
We need to add monitoring to the Postgres db so that we can see how much resources each query is taking.

## How to Test
-'docker compose down --volumes`
- `docker exec semaphore-core python3 ./tools/migrate_db.py`
- Check pg admin to see if the extension has been added:
Screenshot 2025-12-14 215852

- Replace POSTGRES_USER and POSTGRES_DB with real .env variables to access the db container: `docker compose exec db psql -U "POSTGRES_USER " -d "POSTGRES_DB "`
- Run command: `SHOW shared_preload_libraries;`
- The result should be "pg_stat_statements"
- Wait for traffic to run for a bit then query the statements to see what has been logged: 
```
SELECT
    query,
    calls,
    total_exec_time,
    mean_exec_time,
    max_exec_time
FROM pg_stat_statements
ORDER BY total_exec_time DESC
LIMIT 5;
```

- Change the target version to 3.3 and repeat 1 and 2.
- Check extensions and ensure that pg_stat_statements is gone

Note: Rollback will just remove the extension. Statistics will still be gathered, they just won't be able to be queried through the db.

image

### DIFF
--- a/data/dspec/MLP-OP.json
+++ b/data/dspec/MLP-OP.json
@@ -5,7 +5,7 @@
     "author": "Christian Duff",
     "modelFileName": "./MLP-OP",
     "timingInfo":{
-        "active": true,
+        "active": false,
         "offset": 600,
         "interval": 10800
 
--- a/src/API/apiDriver.py
+++ b/src/API/apiDriver.py
@@ -87,15 +87,20 @@ async def get_input(source: str, series: str, location: str, fromDateTime: str, 
         location, 
         datum
     )
+    
+    nowTime = datetime.now(timezone.utc)
+    fromTimeIsInPast = from_time_in_past(fromDateTime, nowTime)
 
     timeDescription = TimeDescription(
-        fromDateTime, 
+        fromDateTime,
         toDateTime,
-        interval
+        interval,
+        stalenessOffset=(None if fromTimeIsInPast else timedelta(hours=7))
     )
 
     provider = SeriesProvider()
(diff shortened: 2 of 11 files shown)


lovelysandlonelys (approved):
Everything works as intended!
image
I wonder if it might be worth adding some documentation (maybe to the read me?) that documents the query and explains what the output means?

It also looks like there are some merge conflicts shouldn't this be migration 3.5?

lovelysandlonelys (changes requested):
It looks like you just have to update the version to 3_5 and correct some documentation, sorry to catch this slightly late, only realized after I returned to main that 3_4 already existed. I do think adding to the readme to explain the query is also a good idea, or maybe just a wiki page, up to your preference I suppose.

abeasley1722:
> Going to need to merge main to get the current updates into this branch. The way we have our version index set up in the migrate file, the update looks for 3.4 which isn't here so 3.5 is never loaded
> 
> Image Image

This can be changed through the query.  SELECT
    LEFT(query, 200) || '…' AS query_preview,
    query AS full_query,
    calls,
    total_exec_time,
    mean_exec_time,
    max_exec_time
FROM pg_stat_statements
ORDER BY total_exec_time DESC
LIMIT 5;

CJQuintero61 (approved):
Works as intended and rollbacks work as well

lovelysandlonelys (approved):
nice documentation and works as intended

PR #815, opened 2025-12-17
Remove old logic from fetch_oldest_generated_time

CJQuintero61:
# What was changed
* The old logic that returned a bool for staleness is removed and the data types returned are correctly fixed to return either None or DateTime
* updated the logging statements in date range validation to use log_error instead of log(). This was causing staleness errors to not be printed when using log( ) and since every validation check that returns false here will cause a semaphore data exception, I changed them to log_error()

# To test
`docker compose up --build -d`
`docker exec semaphore-core python3 -m pytest` See notes below about this
`docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_120hr.json`
`docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Inundation/June/ar_inundation_jun_24h.json`
`docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ThermalRefuge.json`

# Notes
* I tested using the 3 models listed above and running the pytests
* These 7 tests will fail and that is expected. The reason is because of our new query statements using DISTINCT ON is unique to Postgresql and these 7 tests use sqlite to create an in memory db. This fails because sqlite doesn't recognize the "ON" clause causing them to fail.
image

### DIFF
--- a/src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
+++ b/src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
@@ -481,8 +481,6 @@ class SQLAlchemyORM_Postgres(ISeriesStorage):
         if not tupleishResult or not tupleishResult[0][0]: 
             return None
         
-        print(f'Tuplish Result: {tupleishResult}')
-        
         oldestGeneratedTime = pd.to_datetime(tupleishResult[0][0]).tz_localize(timezone.utc)
 
         return oldestGeneratedTime
--- a/src/DataValidation/DataValidationClasses/DateRangeValidation.py
+++ b/src/DataValidation/DataValidationClasses/DateRangeValidation.py
@@ -14,7 +14,7 @@ determines the expected date range via the timeDescription and checks for missin
 #Imports
 from DataClasses import Series
 from DataValidation.IDataValidation import IDataValidation
-from utility import log
+from utility import log_error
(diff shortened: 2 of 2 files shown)


src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
```diff
@@ -477,15 +477,13 @@ def fetch_oldest_generated_time(self, seriesDescription: SeriesDescription, time
         
         tupleishResult = self.__dbSelection(query_stmt).fetchall()
         
-        if not tupleishResult or not tupleishResult[0]: #Data is not present in the DB
-            return False
+        # Data is not present in the DB
+        if not tupleishResult or not tupleishResult[0]: 
+            return None
         
         oldestGeneratedTime = pd.to_datetime(tupleishResult[0]).tz_localize(timezone.utc)
```
abeasley1722: Change all uses of tupleishResult[0] to tupleishResult[0][0] in this function. SO that just the date value is returned and not the row.

abeasley1722 (approved):

src/SeriesStorage/SS_Classes/SQLAlchemyORM_Postgres.py
CJQuintero61: added suggested changes

PR #817, opened 2025-12-17
change log statements in date range validation

CJQuintero61:
Small bug fix to make log statements to log_error() statements so that they always print when data validation fails.

### To Test
`docker compose up --build -d`
`docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_120hr.json` as of 12/16/2025 ndfd_json is failing with stale data. Simply check that the 'is stale' and that the 'time difference: ...' lines are shown 
image


CJQuintero61:
This will be addressed in #815

PR #827, opened 2025-12-27
Fix ingestion issues

CJQuintero61:
### Problem
Cold stunning models are failing to ingest data properly. This is because of this difference check that is tied to the verified time ingestion check using [ now - greatest acquired time ]. This would cause 1 missing data point causing the MRE models to consistently fail instead of trying to ingest for that last data point. The bug that is occurring is that this check compares now (the models run 20 minutes past the hour) to the greatest acquired time which is acquired at the top of the hour, causing this check to always have a difference of 20 minutes and always be below the threshold. 

### Solution 
I removed this check to allow models to try to ingest whenever they are missing data. This allows date range validation to stop the model from running if it is either missing data or is using stale data when models makes it to that point where we want to fail if either of those are the case. If we do want some check to limit how often we go to the ingestion source, we need to think of a new approach since we cannot use NOW and as pointed out in one of Sav's tickets, we cannot use reference time either. 

### To test
2. `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/MRE_Bird-Island_Water-Temperature_120hr.json`
3. `docker exec semaphore-core python3 -m pytest`

### How I tested
I ran every MRE model and regular cold stunning model at different hours. Using verbose logging, the MRE models ingest data properly and have ran successfully every time for me. The regular cold stunning models are a hit and a miss when I ran them. They also ingest properly now but fail due to stale data from ndfd json, and other times they pass. For the purposes of this PR, that is beyond scope and will have to be its own fix/investigation. Also see the ticket for some more information if needed https://github.com/orgs/conrad-blucher-institute/projects/17/views/14?pane=issue&itemId=146879855&issue=conrad-blucher-institute%7Csemaphore%7C826.

### DIFF
--- a/src/SeriesProvider/SeriesProvider.py
+++ b/src/SeriesProvider/SeriesProvider.py
@@ -209,7 +209,7 @@ class SeriesProvider():
         - The max verified time is >= requested toDateTime
 
         NOTE::
-        The now time and toDateTime are both converted to tz naive for comparison.
+        The requested toDateTime is converted to tz naive for comparison.
         """
 
         # get the row with the max verified time in the requested range


src/SeriesProvider/SeriesProvider.py
```diff
@@ -195,23 +195,21 @@ def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescr
...
-            AND the time since acquisition (now - acquired_time) is greater than or equal to the threshold (>= threshold)
-    
-        Returns False (should NOT ingest) if:
-        - The max verified time >= requested toDateTime
-            OR the time since acquisition (now - acquired_time) is less than the threshold (< threshold)
+        True: The database contains all verified times up to the requested toDateTime.
+        Therefore, no new data ingestion is needed.
+
+        False: The database is missing 1 or more verified times up to the requested toDateTime.
+        Therefore, new data ingestion should be ingested.
```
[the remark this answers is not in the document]
CJQuintero61: fixed

src/SeriesProvider/SeriesProvider.py
```diff
@@ -221,23 +219,12 @@ def __check_verified_time_for_ingestion(self, seriesDescription: SeriesDescripti
         if not max_verified_time_row:
             return True
```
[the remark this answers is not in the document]
CJQuintero61: fixed

src/SeriesProvider/SeriesProvider.py
```diff
@@ -221,23 +219,12 @@ def __check_verified_time_for_ingestion(self, seriesDescription: SeriesDescripti
...
         
-        # extract times from the row and add timezone info
+        # extract verified time from the row and add timezone info
```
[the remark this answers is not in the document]
CJQuintero61: fixed

src/SeriesProvider/SeriesProvider.py
```diff
@@ -195,49 +195,36 @@ def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescr
...
         """
 
         # get the row with the max verified time in the requested range
         max_verified_time_row = self.seriesStorage.fetch_row_with_max_verified_time_in_range(seriesDescription, timeDescription)
 
-        # check if we have any rows for the provided series and time description
+        # no rows found means verified times are missing so ingestion is needed.
         if not max_verified_time_row:
-            return True
+            return False
```
abeasley1722: I can see setting it to True causing a lot of problems. Good catch!

src/SeriesProvider/SeriesProvider.py
```diff
@@ -74,8 +74,8 @@ def request_input(self, seriesDescription: SeriesDescription, timeDescription: T
 
         # if we haven't ingested due to staleness, check verified time ingestion
         if not already_ingested_data:
-            should_ingest_for_verified_time = self.__check_verified_time_for_ingestion(seriesDescription, timeDescription)
-            if should_ingest_for_verified_time:
+            all_verified_times_present = self.__check_verified_time_for_ingestion(seriesDescription, timeDescription)
```
ccftissot: Are we really checking that all verified times are present?  if not, then the new variable name is misleading

ccftissot: are we really checking that all verified times are present?

src/SeriesProvider/SeriesProvider.py
```diff
@@ -195,49 +195,36 @@ def __data_ingestion_query(self, seriesDescription: SeriesDescription, timeDescr
...
-        - The max verified time < requested toDateTime (more data might be available)
-            AND the time since acquisition (now - acquired_time) is greater than or equal to the threshold (>= threshold)
-    
-        Returns False (should NOT ingest) if:
-        - The max verified time >= requested toDateTime
-            OR the time since acquisition (now - acquired_time) is less than the threshold (< threshold)
+        True: The database contains all verified times up to the requested toDateTime.
+        Therefore, no new data ingestion is needed.
+
+        False: The database is missing 1 or more verified times up to the requested toDateTime.
```
ccftissot: not sure this is an accurate statement

src/SeriesProvider/SeriesProvider.py
CJQuintero61: reverted back

src/SeriesProvider/SeriesProvider.py
CJQuintero61: reverted back

src/SeriesProvider/SeriesProvider.py
```diff
@@ -203,12 +203,10 @@ def __check_verified_time_for_ingestion(self, seriesDescription: SeriesDescripti
...
-        Returns False (should NOT ingest) if:
-        - The max verified time >= requested toDateTime
-            OR the time since acquisition (now - acquired_time) is less than the threshold (< threshold)
+        - The max verified time is < requested toDateTime (more data might be available)
+
+        False (should NOT ingest) if:
+        - The max verified time is >= requested toDateTime
 
         NOTE::
         The now time and toDateTime are both converted to tz naive for comparison.
```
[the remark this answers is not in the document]
CJQuintero61: fixed

CJQuintero61:
We decided to remove this check and proceed with this PR's changes. The reasoning was that it is better to ask too often than to not ask enough

PR #847, opened 2026-01-21
835 select latest output endpoint causing high cpu usage

matdenkas:
## What:
- Output latest endpoint updated to handle 422 Unprocessable Entity if no models were provided
- ORM output latest method's queries updated to be explicit
- Orm output latest method's queries updated to have limit one.

## Why:
The output latest endpoint is the main endpoint flare uses to retrieve predictions. It can take a massive amount of time to run when quering MRE data due to the un-optimized queries. Adding limit one with the correct order massively improves the query time.

## Test:
1) Build and run
2) Run some models
3) Use the output latest endpoint at `https://localhost:8888/docs`


lovelysandlonelys (approved):
works as expected, when no model name is provided I see the error message added and returns both MRE and non MRE model results as it should. nice work!

PR #862, opened 2026-02-05
increasing staleness offset for NDFD_JSON and SEMAPHORE

lovelysandlonelys:
Increasing staleness offset according to some findings in Dev and Prod! I increased all of the NDFD sources to be 11 hours rather than 10 because we were still having occasional failures due to 10hr 30min staleness. I also increased the staleness for the magnolia transform models so that their SEMAPHORE source has a 13hr staleness matching since we need 12 hours of the past predictions from the first magnolia model in order for the transform models to run.


abeasley1722 (approved):
Woo, thats a lot of files! Great catch. Works as intended.

PR #904, opened 2026-03-16
Monthly archive of the inputs table

### LINKED ISSUE #869: Monthly Archiving of Inputs Table
Right now the inputs endpoint has priority in Flare calls and the calls have to sort through the ENTIRE giant inputs table. If we archive the table every month we limit the number of rows in the table that are being sorted through.

abeasley1722:
Right now the inputs endpoint has priority in Flare calls and the calls have to sort through the ENTIRE giant inputs table. If we archive the table every month we limit the number of rows in the table that are being sorted through.

I plan to create a new inputs table partitioned by month. Old monthly partitions will be detached, and a cron script will run each month to create new partitions and detach old ones.


ccftissot:
note that we probably don't want to partition by months because we want to be able to have the data we need in the main partition at any time.  If we partition based on months, then on the 1st we would need to go fetch the data from the detached partitions to run current model.  Instead, we should have only one other archived partition and have data older than a number of days moved to that partition on a schedule.

PR #909, opened 2026-03-25
Add multi member runs to model runner

### LINKED ISSUE #853: Multi Member Runs In Model Runner
In model runner we need to add a loop for if a DSPEC is for an ensemble model (like CRPS) we look over the members and then stack the predictions before handing them over to the output handler. This ticket would also encompass checking that output handler is given the correct 3d shape.

abeasley1722:
## Objective
Enable multi-member model execution in `ModelRunner` by:
- Ensuring the number of loaded models matches the number of members defined in the DSPEC  
- Iterating through each model to generate and stack prediction outputs  

---

### DSPEC Parser
- Added an optional field `modelFileNamePattern` to support multi-model configurations  
- Usage:
  - `modelFileName` → for single-model DSPECs  
  - `modelFileNamePattern` → for multi-model DSPECs  

### ModelRunner
- Updated logic to support both single-model and multi-model runs
- Implemented loop over multiple models and stacking of prediction outputs  

---

## How to Test
```bash
docker compose up -d --build
docker exec semaphore-core python -m pytest -s
```
## Note:
Feel free to run models, but it will fail at output handler since it expects 2d arrays at the moment. I added print statements before output handler to confirm the correct output.  

Results from print statements in model runner:

# Test Results
Screenshot 2026-03-25 100030
Screenshot 2026-03-25 100045

# 24 hr Cold stunning models
Screenshot 2026-03-25 092400
Screenshot 2026-03-25 092444

### DIFF
--- a/src/ModelExecution/modelRunner.py
+++ b/src/ModelExecution/modelRunner.py
@@ -58,10 +58,9 @@ class ModelRunner:
 
         # Use first model for shape (all models should match)
         expectedShape = models[0].input_shape
-        first_shape = models[0].input_shape
         
         for m in models:
-            if m.input_shape != first_shape:
+            if m.input_shape != expectedShape:
                 raise Semaphore_Exception("Model input shapes do not match")
             
         expectedShape = (len(input_vectors),) + expectedShape[1:]
--- a/src/ModelExecution/modelRunner.py
+++ b/src/ModelExecution/modelRunner.py
@@ -17,61 +17,120 @@ from DataClasses import SemaphoreSeriesDescription, Series
 from utility import log, construct_true_path
 from exceptions import Semaphore_Exception
 from ModelExecution.dspecParser import Dspec
-
+import re
 import datetime
 from os import path, getenv
 from numpy import reshape
+import numpy as np
+import glob
 from tensorflow.keras.models import load_model
 
 
 class ModelRunner:
-     
 
     def make_predictions(self, DSPEC: Dspec, input_vectors: list[any], reference_time: datetime) -> Series:
-        """Computes a prediction but does not save it to persistent storage"""
+        """
+        This function runs predictions using one or multiple models and processes the results.
 
-        log('Init load model....')
-        model = self.__load_model(DSPEC.modelFileName)
+        - Loads model(s) based on the DSPEC configuration
+        - Reshapes input vectors to match the expected model input shape
+        - Runs predictions for each model using the same input data
+        - Stacks all model predictions into a single 3D array 
+        (models, input_vectors, outputs)
+        - Applies post-processing using the specified output handler
+        - Wraps the processed results into a Series object with metadata
+        - Returns the final Series containing the prediction results
+        """
+        log('Init load model(s)....')
+
+        models = self.__load_models(DSPEC)
 
         log('Init shape inputs....')
-        # The input vector(s) must be reshaped to match the model's input shape
-        # The first dimension of the expected shape is left None by TensorFlow 
-        # for us to fill in the amount of input vectors we are providing.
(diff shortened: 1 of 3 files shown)


src/ModelExecution/modelRunner.py
```diff
@@ -17,61 +17,129 @@
...
-        log('Init compute prediction....')
-        prediction =  model.predict(shapedInputs) 
+        # Use first model for shape (all models should match)
+        expectedShape = models[0].input_shape
+        first_shape = models[0].input_shape
+        
+        for m in models:
+            if m.input_shape != first_shape:
+                raise Semaphore_Exception("Model input shapes do not match")
+
```
CJQuintero61, line 65: I think this could be cleaned up by placing this check inside the other for loop before calling model.predict.

src/tests/UnitTests/test_modelRunner.py
```diff
@@ -15,112 +15,231 @@
...
+    with patch('src.ModelExecution.modelRunner.output_handler_factory',
+               mock_outputHandlerFactory(OH_MOCK)):
+
+        with patch.object(ModelRunner, "_ModelRunner__load_models") as mock_loader:
+
+            num_models = 3 if multi else 1
+            mock_loader.return_value = mock_models_with_order(num_models)
+
+            runner = ModelRunner()
+            result_series = runner.make_predictions(
```
CJQuintero61, line 153: This unit test fails for 5 tests because of runner.make_predictions() for the reason below.        

models = self.__load_models(DSPEC)
>       expectedMemberCount = DSPEC.outputInfo.expectedOutputShape.memberCount
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: 'NoneType' object has no attribute 'memberCount'

src/ModelExecution/modelRunner.py
```diff
@@ -17,61 +17,129 @@
...
-        # for us to fill in the amount of input vectors we are providing.
-        expectedShape = model.input_shape
-        expectedShape = (len(input_vectors),) + expectedShape[1:] 
-        shapedInputs = reshape(input_vectors, expectedShape) 
 
-        log('Init compute prediction....')
-        prediction =  model.predict(shapedInputs) 
+        # Use first model for shape (all models should match)
+        expectedShape = models[0].input_shape
+        first_shape = models[0].input_shape
```
CJQuintero61: first_shape can be removed since its only used on line 64 so instead just check if m.input_shape != expectedShape:

CJQuintero61 (changes requested):
Image
------------------------------------------
Image

Just fix the unit tests and small cleanups. The core logic in model runner looks good to me and I was able to add print statements to confirm shapes (they crashed as expected in output handler). I think the dspecParser changes are also fine and using modelFilePattern is fine as a way to distinguish multi member models to single member models.

src/ModelExecution/modelRunner.py
abeasley1722: I think it’s better to check all the model shapes first, so we don’t start running predictions and then fail halfway through. Unless you mean placing the entire for loop inside the other for loop, but I don't see the point in that version.

src/tests/UnitTests/test_modelRunner.py
abeasley1722: Thankss. I made a change and forgot to rebuild the containers.

CJQuintero61 (approved):
Thanks for the fixes. Pytests pass and the changes are fine with me.

CJQuintero61:
merging to crps-implementation to continue working on crps testing. The changes will be combined into 1 PR.

PR #916, opened 2026-03-30
Fixing 500 when no data is returned

### LINKED ISSUE #915: 500 when requesting Magnolia model from Semaphore-dev API
Image

The issue has to do with NoneType handling, either caused by missing data or another bug

matdenkas:
## What?
The serializer was causing a 500 when it received a series with dataFrame set to None
series.dataFrame was being set to None when no data was found for a query.

## Fix:
I added a case to the serializer for is the series exists but the data frame doesnt, it serializes an empty data field.

## Test
2. curl curl http://localhost:8888/output/modelName=ar_inundation_mar_24h/modelVersion=1.0.0/series=pInundation/location=Aransas/fromDateTime=2026033102/toDateTime=2026033113?interval=3600
3. See no 500

### DIFF
--- a/src/API/apiDriver.py
+++ b/src/API/apiDriver.py
@@ -234,6 +234,13 @@ def serialize_series(series: Series) -> dict[any]:
 
     if series is None:
         return dict()
+    
+    if series.dataFrame is None:
+        # In this case we have no data to serialize but we still want to return the description and time description
+        serialized = jsonable_encoder(series, exclude={'_Series__dataFrame'})
+        serialized['isComplete'] = True # Ensure isComplete always exists for backward compatibility
+        serialized['_Series__data'] = [] # Ensure data always exists for backward compatibility
+        return serialized
 
     if isinstance(series.description, SemaphoreSeriesDescription):
         return serialize_output_series(series)


lovelysandlonelys (approved):
small change that increases readability of API responses 👍🏻

CJQuintero61 (approved):
Good

src/API/apiDriver.py
```diff
@@ -234,6 +234,13 @@ def serialize_series(series: Series) -> dict[any]:
 
     if series is None:
         return dict()
+    
+    if series.dataFrame is None:
+        # In this case we have no data to serialize but we still want to return the description and time description
+        serialized = jsonable_encoder(series, exclude={'_Series__dataFrame'})
+        serialized['isComplete'] = True # Ensure isComplete always exists for backward compatibility
```
ccftissot, line 241: hmm ... that should be false, no? surely if we didn't get anything, then the data is not complete?

src/API/apiDriver.py
```diff
@@ -234,6 +234,13 @@ def serialize_series(series: Series) -> dict[any]:
 
     if series is None:
```
ccftissot, line 235: that does not look right? if we didn't get a series at all, something bad happened? shouldn't we return some meaning full http code?  sounds like we need a ticket to change our API to return meaningful code.  This code needs to be refactored too as the execution control and responsibility assignment across functions is all over the place.  however this is out of scope for this ticket

ccftissot (changes requested):

src/API/apiDriver.py
matdenkas: Yes, as the comment says, its for backwards compatibility. Flare warns if that flag is false, ive no clue what CBOCP does, but this is the prior behavior of the API. I can write a note on the API improvment ticket

src/API/apiDriver.py
matdenkas: Agreed

matdenkas: Ill write this as a note on the improvment ticket

src/API/apiDriver.py
matdenkas: I want to be clear if you want me to change it I have no issues doing that.

ccftissot: as we discussed let's leave it as is for now.  fix in a later ticket

ccftissot (approved):

PR #927, opened 2026-03-31
adding interpolation to semaphore series

### LINKED ISSUE #921: Add Interpolation to Magnolia Transform Models
Right now, if the base Magnolia model has 1 failure the Magnolia modes have 12 because they use 12 hours of past data and that one missing point slides through the window. If we add interpolation to the Semaphore series in the Magnolia Transform models we'll reduce the cascade from 1:12 to a 1:1 relationship.

lovelysandlonelys:
A small change to add interpolation to the Semaphore series that the magnolia transform models ingests. 

There isn't a great way to test this but what I did was run the base magnolia model once and then the transform to see that interpolation kicked off for that series.

image

Test: 

1. build containers `docker compose up -d --build`
2. run base magnolia model `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Magnolia/12/magnolia_12.json -v`
3. run transform `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Magnolia/12/magnolia_transform_12.json -v`

### DIFF
--- a/data/dspec/Magnolia/12/magnolia_transform_12.json
+++ b/data/dspec/Magnolia/12/magnolia_transform_12.json
@@ -27,7 +27,15 @@
             "interval": 3600,
             "range":  [7, -5],
             "outKey": "MG_12_output_12",
-            "stalenessOffset": 45000
+            "stalenessOffset": 45000,
+            "dataIntegrityCall": {
+                "call": "PandasInterpolation",
+                "args": {
+                    "limit": "3600",
+                    "method": "linear",
+                    "limit_area": "inside"
+                }
+            }
         },
         { 
             "_name": "Wind Direction Prediction",
--- a/data/dspec/Magnolia/24/magnolia_transform_24.json
+++ b/data/dspec/Magnolia/24/magnolia_transform_24.json
@@ -27,7 +27,15 @@
             "interval": 3600,
             "range":  [19, 7],
             "outKey": "MG_24_output_12",
-            "stalenessOffset": 45000
+            "stalenessOffset": 45000,
+            "dataIntegrityCall": {
+                "call": "PandasInterpolation",
+                "args": {
+                    "limit": "3600",
+                    "method": "linear",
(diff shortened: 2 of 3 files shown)


data/dspec/Magnolia/12/magnolia_transform_12.json
```diff
@@ -27,7 +27,15 @@
             "interval": 3600,
             "range":  [7, -5],
             "outKey": "MG_12_output_12",
-            "stalenessOffset": 45000
+            "stalenessOffset": 45000,
+            "dataIntegrityCall": {
+                "call": "PandasInterpolation",
+                "args": {
+                    "limit": "3600",
```
ccftissot, line 34: do we want to allow for more missing data for interpolation?  this would only allow one hour of missing data,  or am i reading this wrong? same comment for all 3 transforms.

ccftissot (approved):
Looks good!  ran all 3 transform models and interpolation is kicking in on all 3.  just have one question around how much gap should we allow.  Philippe says no more than 2 hours so if 1 hour is not enough we can always increase to 2.

ps:  and yes i finally got semaphore to run in containers on my machine:  woohoo, i can actually test things now.

data/dspec/Magnolia/12/magnolia_transform_12.json
lovelysandlonelys: I think it would be worth pushing and then going forward from there. In the meeting we were talking about the case we usually see, which is one missing point that slides through the entire time series, so although allowing for multi-point interpolation is allowed I think it would be interesting to see if we even run in that at all before raising the threshold?

abeasley1722 (approved):
Works as intended! Great work

data/dspec/Magnolia/12/magnolia_transform_12.json
ccftissot: i agree

PR #933, opened 2026-04-07
date range clipping in validation according to model needs

### LINKED ISSUE #919: Lighthouse Haunted Data Gap Fix

## Current Behavior

## Fix

lovelysandlonelys:
## Lighthouse Haunted Data Gap Fix #919

**What was changed**

This PR fixes the "haunted gap" failures we were seeing on Bird Island and any other model that over-requests data to support interpolation buffering. `DateRangeValidation` was validating the entire requested window including buffer slots that the model never actually reads. When a Lighthouse sensor dropout landed exactly on `fromDateTime` (the buffer slot), it would fail validation even though the model would have run fine without that point. The fix passes `vectorOrder` indexes into the validation step so `DateRangeValidation` only checks the slots the model actually reads. The original series is never mutated.

Changes:
- `dataGatherer.py` — builds a `{outKey: (min_index, max_index)}` lookup from `dspec.orderedVector` at the top of `get_data_repository` and threads it down to `__validate_series`
- `DateRangeValidation.py` — accepts optional `indexes: tuple[int, int]`; if provided, clips the dataframe to model-read rows only before building the expected index. Error logging now fires next to the failure and includes the validated window bounds and whether indexes were applied.

**To Test**

**Other things to validate**

- Confirm the vector builder logs `amnt_found: 27, indexed_len: 25` for the Bird Island water/air temp series — this means 27 rows were fetched, the buffer was clipped, and 25 were validated
- Confirm models without `vectorOrder` entries still run normally 
- If you think there was a better way to do this.... let me know.

### DIFF
--- a/src/DataValidation/DataValidationClasses/DateRangeValidation.py
+++ b/src/DataValidation/DataValidationClasses/DateRangeValidation.py
@@ -46,36 +46,20 @@ class DateRangeValidation(IDataValidation):
         # The original must stay intact — buffer slots outside the indexed window
         # are still valid data we want to store.
         df_to_validate = series.dataFrame.copy()
-
-        # --- Clipping ---
-        # If vectorOrder indexes were provided, clip the dataframe to only the rows
-        # the model actually reads. Buffer slots outside this window are intentionally
-        # excluded — their absence should not cause validation to fail.
+        
         if self.indexes is not None:
-            # Unpack: min_index is the closest-to-now slot (right/future side),
-            # max_index is the furthest-from-now slot (left/past side).
             min_index, max_index = self.indexes
+            df_to_validate = df_to_validate.iloc[min_index:max_index]
 
-            # The dataframe is ordered oldest → newest after the reindex in DataGatherer.
-            # -(max_index + 1) counts back from the end to include the furthest past slot.
-            # e.g. max_index=25 → iloc[-26:] keeps the 26 model-read rows from the left.
-            clip_start = -(max_index + 1)
-
-            # len(df) - min_index trims future buffer rows from the right.
-            # When min_index=0 this evaluates to len(df), i.e. no right-side trim.
-            # `or None` converts 0 → None so iloc doesn't produce an empty slice
-            # (iloc[x:0] is empty, iloc[x:None] goes to the end — which is what we want).
-            clip_end = len(df_to_validate) - min_index or None
-
-            df_to_validate = df_to_validate.iloc[clip_start:clip_end]
+            if len(df_to_validate) == 0:
+                log_error(f'DateRangeValidation: indexes={self.indexes} produced an empty clipped window for a dataframe of length {len(series.dataFrame)}')
+                return False
 
         # Set timeVerified as the index so we can reindex against an expected date range.
         df_to_validate.set_index('timeVerified', inplace=True)
-        
-        # --- Expected index ---
-        # When indexes are provided, bounds come from the clipped dataframe's actual
-        # timestamps — using timeDescription here would re-expand back to the full window.
-        # When indexes is None, use timeDescription bounds (original behavior) so that
+
+        # When indexes are provided use the clipped frame's actual bounds.
+        # When indexes is None use timeDescription bounds (original behavior) so
         # dataframes missing values at the edges relative to timeDescription still fail.
         if self.indexes is not None:
             expected_index = date_range(
@@ -89,7 +73,7 @@ class DateRangeValidation(IDataValidation):
                 end=series.timeDescription.toDateTime,
                 freq=timedelta(seconds=series.timeDescription.interval.total_seconds())
             )
-
+            
         # Reindex to the expected range — any genuinely missing interior timestamps
         # will become NaN rows, which we catch below.
         df_to_validate = df_to_validate.reindex(expected_index)
--- a/src/DataValidation/DataValidationClasses/DateRangeValidation.py
+++ b/src/DataValidation/DataValidationClasses/DateRangeValidation.py
@@ -21,8 +21,16 @@ from pandas import date_range
 
 class DateRangeValidation(IDataValidation):
 
-    def __init__(self, referenceTime: datetime = None):
+    def __init__(self, referenceTime: datetime = None, indexes: tuple[int, int] = None):
+        # referenceTime is used for the staleness check further below.
+        # It is None in unit tests and for series that don't need a staleness check.
         self.referenceTime = referenceTime
+
+        # indexes is the (min_index, max_index) from vectorOrder — the actual slots
+        # the model reads. e.g. (0, 25) for a 26-point model with a left buffer,
+        # or (1, 25) if there is also a right-side buffer.
+        # None means no buffer info was provided; validate the full window as-is.
+        self.indexes = indexes  
         
     def validate(self, series: Series) -> bool:
         """ This method checks for missing date ranges in the the expected time series. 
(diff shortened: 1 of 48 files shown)


src/ModelExecution/dataGatherer.py
```diff
@@ -56,16 +56,26 @@ def get_data_repository(self, dspec: Dspec, referenceTime: datetime) -> dict[str
...
+        dependent_data_repository = self.__request_dependent_data(dependentSeries, referenceTime, vector_index_lookup)
 
         # Call post processing 
         post_processed_series_repository = self.__post_process_data(dependent_data_repository, postProcessCalls)
 
         return post_processed_series_repository
     
 
-    def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], referenceTime: datetime) -> dict[str, Series]:
+    def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], referenceTime: datetime, vector_index_lookup: dict[str, tuple[int, int]] = {}) -> dict[str, Series]:
```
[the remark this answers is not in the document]
lovelysandlonelys: Done.

src/DataValidation/DataValidationClasses/DateRangeValidation.py
```diff
@@ -34,32 +42,79 @@ def validate(self, series: Series) -> bool:
...
+            # e.g. max_index=25 → iloc[-26:] keeps the 26 model-read rows from the left.
+            clip_start = -(max_index + 1)
+
+            # len(df) - min_index trims future buffer rows from the right.
+            # When min_index=0 this evaluates to len(df), i.e. no right-side trim.
+            # `or None` converts 0 → None so iloc doesn't produce an empty slice
+            # (iloc[x:0] is empty, iloc[x:None] goes to the end — which is what we want).
+            clip_end = len(df_to_validate) - min_index or None
+
+            df_to_validate = df_to_validate.iloc[clip_start:clip_end]
```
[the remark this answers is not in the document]
lovelysandlonelys: Ahhhhh i made it more complicated than it needs to be.

src/DataValidation/DataValidationClasses/DateRangeValidation.py
```diff
@@ -21,8 +21,16 @@
...
+    def __init__(self, referenceTime: datetime = None, indexes: tuple[int, int] = None):
+        # referenceTime is used for the staleness check further below.
+        # It is None in unit tests and for series that don't need a staleness check.
         self.referenceTime = referenceTime
+
+        # indexes is the (min_index, max_index) from vectorOrder — the actual slots
+        # the model reads. e.g. (0, 25) for a 26-point model with a left buffer,
+        # or (1, 25) if there is also a right-side buffer.
+        # None means no buffer info was provided; validate the full window as-is.
+        self.indexes = indexes
```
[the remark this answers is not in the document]
lovelysandlonelys: Done.

lovelysandlonelys:
image
Okay so!!! In the midst of trying to write tests for this I was double checking the inputVectorOrder functionality and I think!! I have been indexing things wrong. I think that the series I used to base my indexing off of was a prediction series but the lighthouse series we get for bird island are measurements, so I'm also going to add fixing.... that huge mistake to this PR. If anyone wants to triple check this for me I would appreciate it!

lovelysandlonelys:
when indexing is set to 2, 27 the following points are used, which i'm under the impression is correct because for cold stunning we use the past 24 hours and the now time for water temp mesurments.

image

src/DataValidation/DataValidationClasses/DateRangeValidation.py
```diff
@@ -21,8 +21,16 @@
...
         self.referenceTime = referenceTime
+
+        # indexes is the (min_index, max_index) from vectorOrder — the actual slots
+        # the model reads. e.g. (0, 25) for a 26-point model with a left buffer,
+        # or (1, 25) if there is also a right-side buffer.
+        # None means no buffer info was provided; validate the full window as-is.
+        self.indexes = indexes  
         
     def validate(self, series: Series) -> bool:
         """ This method checks for missing date ranges in the the expected time series.
```
ccftissot, line 36: should we update this documentation?  are we really looking for missing ranges or for missing data points?

src/DataValidation/DataValidationClasses/DateRangeValidation.py
ccftissot, line 41: why less or equal? consider using df.empty to make the test more explicit

ccftissot, line 105: might be worth also logging the min time generated for completeness sake.

src/DataValidation/DataValidationClasses/DateRangeValidation.py
```diff
@@ -34,32 +42,63 @@ def validate(self, series: Series) -> bool:
...
-            end=series.timeDescription.toDateTime, 
-            freq=timedelta(seconds=series.timeDescription.interval.total_seconds())
-        )
+
+        # When indexes are provided use the clipped frame's actual bounds.
+        # When indexes is None use timeDescription bounds (original behavior) so
+        # dataframes missing values at the edges relative to timeDescription still fail.
+        if self.indexes is not None:
+            expected_index = date_range(
+                start=df_to_validate.index[0],
```
ccftissot, line 66: consider changing the code to actually grab the verified time column value of the first and last rows instead of using the index.  This would make the code both more explicit and also less brittle; if for reason the code is changed in the future and the index is set to something else than the verified time, this new code would introduce bugs.

src/ModelExecution/dataGatherer.py
```diff
@@ -56,16 +56,26 @@ def get_data_repository(self, dspec: Dspec, referenceTime: datetime) -> dict[str
         dependentSeries: list[DependentSeries] = dspec.dependentSeries
         postProcessCalls: list[PostProcessCall] = dspec.postProcessCall
 
+        # Build a lookup from outKey → indexes using orderedVector
+        vector_index_lookup: dict[str, tuple[int, int]] = {}
+        ordered_vector = getattr(dspec, 'orderedVector', None)
```
ccftissot, line 61: consider using a more direct approach here, i.e., ordered_vector = dspec.orderedVector.  since we define and initialize our classes properly, we should never have a dspec instance without an orderedVector propery.  I would think if we ever did, we would want to code to fail as that would be a bad thing to happen and we would want an exception to be raised.  To make it clear to readers and find syntactic issues with Pylance, consider doing something like on lines 56 and 57.  so something like ordered_vector: VectorOrder = dspec.orderedVector.

src/ModelExecution/dataGatherer.py
```diff
@@ -56,16 +56,26 @@ def get_data_repository(self, dspec: Dspec, referenceTime: datetime) -> dict[str
...
+        if ordered_vector is not None and hasattr(ordered_vector, 'keys'):
```
ccftissot, line 62: same thing here.  if we have an ordered_vector it must be of type VectorOrder and has to have a "keys" property.

src/ModelExecution/dataGatherer.py
```diff
@@ -56,16 +56,26 @@ def get_data_repository(self, dspec: Dspec, referenceTime: datetime) -> dict[str
...
+            vector_index_lookup = {
+                key: indexes
+                for key, indexes in zip(ordered_vector.keys, ordered_vector.indexes)
+                if indexes != (None, None)  # (None, None) means no buffer info, skip
```
ccftissot, line 66: no need for change here but that is not great that we set the indexes property of the vectorOrder to be (None, None) instead of just None. This is because it implies that we could have values like (None, 12). oh well

src/tests/UnitTests/test_DateRangeValidation.py
```diff
@@ -325,7 +325,62 @@ def test_validate_staleness_fresh_data(self):
...
-        self.assertTrue(validator.validate(series_mock))   
+        self.assertTrue(validator.validate(series_mock)) 
+
+    def test_validate_indexes_missing_value_outside_window_passes(self):
+        """Asserts validate returns True when a missing value falls outside the indexed window.
+
+        The indexed window defines which rows the model actually reads. Any missing values
+        outside that window are buffer slots and should be ignored by validation.
+
+        Setup: 10 rows, indexes=(2, 8) → model reads rows 2-7.
```
ccftissot, line 336: hmm really?? why is index 8 ignored but 2 is included?

src/tests/UnitTests/test_DateRangeValidation.py
```diff
@@ -325,7 +325,62 @@ def test_validate_staleness_fresh_data(self):
...
+        series_mock.timeDescription = MagicMock()
+        series_mock.timeDescription.fromDateTime = times[0]
+        series_mock.timeDescription.toDateTime = times[-1]
+        series_mock.timeDescription.interval = timedelta(hours=1)
+        series_mock.timeDescription.stalenessOffset = None
+
+        validator = DateRangeValidation(indexes=(2, 8))
+        self.assertTrue(validator.validate(series_mock))
+
+    def test_validate_indexes_missing_value_inside_window_fails(self):
```
ccftissot, line 358: consider combining missing values outside with missing valid inside for a more comprehensive test.

ccftissot:
> image Okay so!!! In the midst of trying to write tests for this I was double checking the inputVectorOrder functionality and I think!! I have been indexing things wrong. I think that the series I used to base my indexing off of was a prediction series but the lighthouse series we get for bird island are measurements, so I'm also going to add fixing.... that huge mistake to this PR. If anyone wants to triple check this for me I would appreciate it!

looking at one example, Bird-Island_Water-Temperature_3hr, i see the following:
```json
"dependentSeries": [
        {
            "_name": "Water Temp",
            "location": "SouthBirdIsland",
            "source": "LIGHTHOUSE",
            "series": "dWaterTmp",
            "unit": "celsius",
            "interval": 3600,
            "range": [
                0,
                -26
            ],
            "outKey": "south-bird-island_water-temp_25",
            "dataIntegrityCall": {
                "call": "PandasInterpolation",
                "args": {
                    "limit": "3600",
                    "method": "linear",
                    "limit_area": "inside"
                }
            }
        },
```
then
```json
"vectorOrder": [
        {
            "key": "south-bird-island_water-temp_25",
            "dType": "float",
            "indexes": [
                2,
                27
            ]
        },
```
My understanding is that we are asking to fetch 27 hours of measurements for hours now-27 to now, and then, after they are ordered properly by verified time, take rows at indexes 2 to 27 which would drop hours -27 and -26, and we end up with hours -25 to now for a total of 26 measurements.  it that what we want? looking at the history of these files, it looks like we used to ask for -24 to 0, so I am thinking we are feeding the models too much data and it will use value for hour 0 as the first value of the next dependent data series?  Or am i thinking about this the wrong way?

If we are doing what i think we are doing (vomit) and the indexes property about is meant to include 2 and exclude 27 then wouldn't we be dropping now which would be at index 27? is that what we want to do?

ccftissot (changes requested):
thank you for taking this on Savannah. please let me know if you want to discuss my comments.

lovelysandlonelys:
> > image Okay so!!! In the midst of trying to write tests for this I was double checking the inputVectorOrder functionality and I think!! I have been indexing things wrong. I think that the series I used to base my indexing off of was a prediction series but the lighthouse series we get for bird island are measurements, so I'm also going to add fixing.... that huge mistake to this PR. If anyone wants to triple check this for me I would appreciate it!
> 
> looking at one example, Bird-Island_Water-Temperature_3hr, i see the following:
> 
> ```json
> "dependentSeries": [
>         {
>             "_name": "Water Temp",
>             "location": "SouthBirdIsland",
>             "source": "LIGHTHOUSE",
>             "series": "dWaterTmp",
>             "unit": "celsius",
>             "interval": 3600,
>             "range": [
>                 0,
>                 -26
>             ],
>             "outKey": "south-bird-island_water-temp_25",
>             "dataIntegrityCall": {
>                 "call": "PandasInterpolation",
>                 "args": {
>                     "limit": "3600",
>                     "method": "linear",
>                     "limit_area": "inside"
>                 }
>             }
>         },
> ```
> 
> then
> 
> ```json
> "vectorOrder": [
>         {
>             "key": "south-bird-island_water-temp_25",
>             "dType": "float",
>             "indexes": [
>                 2,
>                 27
>             ]
>         },
> ```
> 
> My understanding is that we are asking to fetch 27 hours of measurements for hours now-27 to now, and then, after they are ordered properly by verified time, take rows at indexes 2 to 27 which would drop hours -27 and -26, and we end up with hours -25 to now for a total of 26 measurements. it that what we want? looking at the history of these files, it looks like we used to ask for -24 to 0, so I am thinking we are feeding the models too much data and it will use value for hour 0 as the first value of the next dependent data series? Or am i thinking about this the wrong way?
> 
> If we are doing what i think we are doing (vomit) and the indexes property about is meant to include 2 and exclude 27 then wouldn't we be dropping now which would be at index 27? is that what we want to do?

From my understanding the models use the past 24 hours of measurements AND and now time measurement, expectating 25 points total, however when looking for this in any of our documentation I couldn't find it, so it  is worth double checking.

src/DataValidation/DataValidationClasses/DateRangeValidation.py
```diff
@@ -34,32 +42,63 @@ def validate(self, series: Series) -> bool:
...
+                end=df_to_validate.index[-1],
+                freq=timedelta(seconds=series.timeDescription.interval.total_seconds())
+            )
```
matdenkas, line 69: I am really late on this review so I am not going to block this. I will bring up separation of concerns. Personally I dont think data validation should include this clipping code, this is a whole other step that doesnt really have to do with validation. This also means that this clipping code would have to be added to any future data validation methods that are written. A cleaner design would be to encapsulate this as a class or a function and have it called as data gatherer. IMO

ccftissot: i actually think that any validation of input data should be done on the clipped range. we could discuss passing the clipped range to the the data validation interface instead of the full range.

lovelysandlonelys:
moved to two different branches for seperation of issues. 

https://github.com/conrad-blucher-institute/semaphore/pull/954

https://github.com/conrad-blucher-institute/semaphore/pull/953

Issue #920, opened 2026-03-30
Overrequest Lighthouse Minute Data


lovelysandlonelys:
85% of lighthouse failures are for the 'now-point', however we request six minute data past the hour, so theoretically there should be points available to interpolate the 'now-point' from if we over request data on the right side of the window. One thing to keep in mind would be ensure we drop the first point from the vector order.

lovelysandlonelys:
The original approach in PR #934 handled now-point estimation inside LIGHTHOUSE.py by extending the fetch window and averaging sub-hourly readings to synthesize the missing point. This was flagged during review for bypassing existing interpolation infrastructure and opaquely targeting only toDateTime in a way invisible to the modeler. 

The approach that was settled on relies on the ticket that validates on clipped data: https://github.com/conrad-blucher-institute/semaphore/pull/954 and just extends lighthouse requests into the future by one hour. So we'll get a range of 26 hours in the past, the now point, and one hour in the future. The validation will work only one what's indexed which should be the past 24 hours of data and the now point.

lovelysandlonelys:
<img width="1440" height="960" alt="Image" src="https://github.com/user-attachments/assets/19b07da8-a011-4a08-b2a9-263697823236" />

PR #934, opened 2026-04-09
using overreturned values to estimate now point and test

lovelysandlonelys:
## Overrequest Lighthouse Minute Data #920

**What was changed**

85% of Lighthouse failures were on the now-point (the most recent hourly slot) which I'm now questioning becasue of [PR#933](https://github.com/conrad-blucher-institute/semaphore/pull/933)'s discovery but figured I might as well see this through anyway . Lighthouse publishes 6-minute data faster than hourly aggregates, so by the time the model runs there are often already a few sub-hourly readings past the hour that we can use to estimate the missing now-point.

The fix is entirely self-contained in `LIGHTHOUSE.py`. The fetch window is quietly extended 1 hour past `toDateTime` to capture any sub-hourly readings that have come in. If the now-point is missing but at least one sub-hourly reading exists in that window, they're averaged and synthesized as the now-point. The rest of the pipeline sees a complete series and has no idea this happened.

Changes:
- `LIGHTHOUSE.py` — extends `toString` by 1 hour, collects six min readings past `toDateTime` into a separate bucket, and synthesizes the now-point from their average if it's absent
- `test_LIGHTHOUSE.py` — adds `test_now_point_estimated_from_sub_hourly_data` which mocks the API response with a missing now-point and two sub-hourly readings, asserting the synthesized value equals their average and doesn't appear in the series as extra rows

**To Test**

2. `docker exec semaphore-core python3 -m pytest src/tests/IntegrationTests/test_LIGHTHOUSE.py`

- Note: dspec `range` and `indexes` are unchanged — this fix lives entirely in ingestion as when looking over the lighthouse class I thought this would be simpler but if we hate it I can change it.

### DIFF
--- a/src/DataIngestion/DI_Classes/LIGHTHOUSE.py
+++ b/src/DataIngestion/DI_Classes/LIGHTHOUSE.py
@@ -81,7 +81,7 @@ class LIGHTHOUSE(IDataIngestion):
         # These are only used to estimate the now-point if it's missing — they
         # never enter the series directly.
         extendedToDateTime = timeDescription.toDateTime + timedelta(hours=1)
-        toString = timeDescription.toDateTime.strftime('%m/%d/%y').replace('/', '%2F')
+        toString = extendedToDateTime.strftime('%m/%d/%y').replace('/', '%2F')
         
         ### Find what LIGHTHOUSE calls the series we want
         # SIM = series info map
--- a/src/DataIngestion/DI_Classes/LIGHTHOUSE.py
+++ b/src/DataIngestion/DI_Classes/LIGHTHOUSE.py
@@ -16,7 +16,7 @@ from DataClasses import Series, SeriesDescription, get_input_dataFrame, TimeDesc
 from DataIngestion.IDataIngestion import IDataIngestion
 from utility import log
 
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timedelta
 from urllib.error import HTTPError
 from urllib.request import urlopen
 import json
@@ -75,6 +75,12 @@ class LIGHTHOUSE(IDataIngestion):
         
         # Reformat and sterilize datetimes
         fromString = timeDescription.fromDateTime.strftime('%m/%d/%y').replace('/', '%2F')
+        
+        # Extend the fetch window 1 hour past toDateTime to capture any 6-minute
+        # readings Lighthouse has published since the last hourly aggregate.
+        # These are only used to estimate the now-point if it's missing — they
+        # never enter the series directly.
+        extendedToDateTime = timeDescription.toDateTime + timedelta(hours=1)
         toString = timeDescription.toDateTime.strftime('%m/%d/%y').replace('/', '%2F')
         
         ### Find what LIGHTHOUSE calls the series we want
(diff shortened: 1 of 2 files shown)


ccftissot:
> ## Overrequest Lighthouse Minute Data #920
> **What was changed**
> 
> 85% of Lighthouse failures were on the now-point (the most recent hourly slot) which I'm now questioning becasue of [PR#933](https://github.com/conrad-blucher-institute/semaphore/pull/933)'s discovery but figured I might as well see this through anyway . Lighthouse publishes 6-minute data faster than hourly aggregates, so by the time the model runs there are often already a few sub-hourly readings past the hour that we can use to estimate the missing now-point.
> 
> The fix is entirely self-contained in `LIGHTHOUSE.py`. The fetch window is quietly extended 1 hour past `toDateTime` to capture any sub-hourly readings that have come in. If the now-point is missing but at least one sub-hourly reading exists in that window, they're averaged and synthesized as the now-point. The rest of the pipeline sees a complete series and has no idea this happened.
> 
> Changes:
> 
> * `LIGHTHOUSE.py` — extends `toString` by 1 hour, collects six min readings past `toDateTime` into a separate bucket, and synthesizes the now-point from their average if it's absent
> * `test_LIGHTHOUSE.py` — adds `test_now_point_estimated_from_sub_hourly_data` which mocks the API response with a missing now-point and two sub-hourly readings, asserting the synthesized value equals their average and doesn't appear in the series as extra rows
> 
> **To Test**
> 
> 1. `docker compose up --build -d`
> 2. `docker exec semaphore-core python3 -m pytest src/tests/IntegrationTests/test_LIGHTHOUSE.py`
> 3. `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_120hr -v`
> 
> * Note: dspec `range` and `indexes` are unchanged — this fix lives entirely in ingestion as when looking over the lighthouse class I thought this would be simpler but if we hate it I can change it.

@lovelysandlonelys :thanks for taking this issue on.  i think the overall approach is interesting, but, before I review the code, I have a couple of comments:
1. within the approach: i don't think we should average the extra points to get the boundary point (as you describe above) .  I think we should interpolate between 2 points on each side of the boundary. We should also be careful to not interpolate over too many missing values.  In this case, we only get one extra hours, so we are probably ok interpolating between the previous known value and the next know value.  However, when interpolating, make sure that we only add the boundary point, not any other point to the ingested data.  This takes me to my other point.
2. I can see a couple of issues with the approach: (1) we are side-stepping any other interpolation that is supported by Semaphore and forcing it in a black box manner from the model and modeler.  I think that is an issue. (2) why only do that for the now? Also, a quick look at the code seems to show that we don't actually do that only for the real now, we do that for the todatetime.  that is problematic in my opinion because it seems a bit arbitrary to do that for only one point in the requested data.

Have we tried extending the window of call to the ingestion class instead of changing the ingestion class?  what happens if we setup a dependent series to request [1, -23] instead of [0,-23]?  does that cause issues in the code? and if so, what kind of issue and can we fix the code to handle this situation gracefully?  This way we wouldn't be bypassing any specs and any explicit can clear features and behaviors of semaphore. 

We need to have that conversation (here or in our meeting) before we move forward with this change.  I'll request a change in my review to block this code moving forward for now.

ccftissot (changes requested):
before doing the code review I would like to discuss the approach.  please see my comment on the PR's conversation thread.

lovelysandlonelys:
reworking approach

PR #947, opened 2026-04-20
940 database migration to add statistics table

### LINKED ISSUE #940: Database migration to add statistics table
# Goal
We need to create the statistics table for MRE and CRPS models.

# Rows
The following rows should be created
- ID
- outputID
- p1
- p10
- p5
- p25
- p50
- p75
- p90
- p95
- p99
- min
- max
- mean
- std_dev

All rows should be NON NULLABLE and of data type float (except for ID and outputID obviously)

# Notes
- Ensure rollbacks and rollforwards work as expected when testing
- ID is the PK entry for that row
- outputID is the FK entry to the raw data row in the outputs table that these statistics are tied to

abeasley1722:
# Goal
We need to create the statistics table for MRE and CRPS models.

# Rows
The following rows should be created
- ID
- outputID
- p1
- p10 - needs to be added
- p5
- p25
- p50
- p75
- p90 - needs to be added
- p95
- p99
- min
- max
- mean
- std_dev

All rows should be NON NULLABLE and of data type float (except for ID and outputID obviously)

# Notes
- Ensure rollbacks and rollforwards work as expected when testing
- ID is the PK entry for that row
- outputID is the FK entry to the raw data row in the outputs table that these statistics are tied to

# Testing
- docker compose down --volumes  
- docker compose up -d --build   
-  docker exec semaphore-core python3 tools/migrate_db.py
- Change the target version to 3.6 and repeat steps 2 and 3


tools/DatabaseMigration/target_version.json
```diff
@@ -1,4 +1,4 @@
 {
-    "Target Version" : "3.6",
+    "Target Version" : "3.7",
     "Description" : "Retires current outputs table and creates the new outputs table."
```
CJQuintero61: Need to update the description for version 3.7 adding the statistics table

tools/DatabaseMigration/3_7/3_7_DatabaseMigration.py
```diff
@@ -0,0 +1,73 @@
...
+            connection.commit()
+            return True
+
+    def rollback(self, databaseEngine: Engine) -> bool:
+        """ Rollback behavior:
+            - Drop statistics table
+        """
+        with databaseEngine.connect() as connection:
+
+            # Drop the NEW outputs table and it's constraints and sequence.
```
CJQuintero61: this comment needs to be updated since it drops the statistics table

CJQuintero61 (approved):
I found a few minor things. The migration created the table and rollbacked properly though. I asked copilot to review as well, so make sure to address those issues as well.

CJQuintero61 (changes requested):
Sorry, meant to block as this is going to main and will require you to address all comments and will make all reviews invalid anyway whenever there are new commits. 

Just make sure to address the previous comments and copilot requests.

tools/DatabaseMigration/3_7/3_7_DatabaseMigration.py
```diff
@@ -0,0 +1,73 @@
...
+            :return: bool indicating successful update
+
+           
+
+        """
+        with databaseEngine.connect() as connection:
+
+            # Create the new statistics table with it's constraints
+            connection.execute(text("""
+                CREATE TABLE public."statistics" (
```
ccftissot: let's call it output_statistics instead of just statistics. In the future, we may have other types of stats such as model run failure and success, etc.

tools/DatabaseMigration/3_7/3_7_DatabaseMigration.py
```diff
@@ -0,0 +1,73 @@
...
+
+                    "min" FLOAT NOT NULL,
+                    "max" FLOAT NOT NULL,
+                    "mean" FLOAT NOT NULL,
+                    "std_dev" FLOAT NOT NULL,
+
+                    CONSTRAINT "fk_statistics_output"
+                        FOREIGN KEY ("outputID")
+                        REFERENCES public."outputs"("id")
+                        ON DELETE CASCADE
```
ccftissot: oh i don't think we want to do that, do we? do we do that for other tables?  this is typically regarded as risky practice because it can (1) silently cause unintended deletions which makes it hard to debug and (2) accidentally increase the scope of some deletion (when one table has an FK to another table that has an FK, etc.) .  It is preferable to explicit delete rows in code rather than let the DB do that behind the scenes.

ccftissot (changes requested):
just a couple of comments. let me know if you want to discuss.  If we created other tables with ON DELETE CASCADE, we need to create a new ticket to do a DB migration to remove these triggers.

ccftissot:
when i  tested, i got an error when trying to run the migrate script. this is because I have a __pycache__ folder on my machine and it got copied over when creating the containers.  This causes the create_version_list method to fail because it can't cast __pycache__  into an int.  We need to update the .dockerignore file to ignore  __pycache__ not just in the root folder but in all folders. So something like **/__pycache__ instead of just __pycache__.  We should probably eventually make that code more robust by having it ignore any folder that is not of the form int_int.  I will create a ticket for that.

Other than that, migration and rollback works as intended.

lovelysandlonelys (dismissed):
rollback and rollforward both work, nothing looks wrong to me. i saw you got rid of an on delete cascade on miss tissots suggestion, i wonder if we should have any other behavior for on delete but not sure so not a blocker in my mind

tools/DatabaseMigration/3_7/3_7_DatabaseMigration.py
```diff
@@ -0,0 +1,73 @@
...
+
+        """
+        with databaseEngine.connect() as connection:
+
+            # Create the new statistics table with it's constraints
+            connection.execute(text("""
+                CREATE TABLE public."statistics" (
+                    "id" INTEGER GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+                    "outputID" INTEGER NOT NULL,
+
```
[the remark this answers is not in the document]
ccftissot: we should add a unique constraint for sure

ccftissot (dismissed):
looks good.  i am not sure about adding an index already as we don't know that it is going to need some performance tuning, but since you have it in there, i would leave it.  I do think that we need a unique constraint:  we should only have at max 1 row per output ID in this table.

ccftissot (approved):
Looks good.

lovelysandlonelys (approved):

PR #954, opened 2026-04-21
919 lighthouse haunted data gap fix electric boogaloo

### LINKED ISSUE #919: Lighthouse Haunted Data Gap Fix

lovelysandlonelys:

## Current Behavior

## Fix

Pass only the indexed data to validation. I made some changes based on feedback from Mrs. Tissot on the previous PR, rather than the clipping being done inside of data validation, we do the clipping outside and then pass the clipped series to data validation. Things should be a written a bit more explicitly as well. I thought about Matthew's idea of having this it's own class but I think that might be overkill for what is ultimately a temporary shift just so that data validation looks at the right points. (If someone would add copilot I would appreciate it.)

**To Test**

### DIFF
--- a/src/DataValidation/DataValidationClasses/DateRangeValidation.py
+++ b/src/DataValidation/DataValidationClasses/DateRangeValidation.py
@@ -25,12 +25,11 @@ class DateRangeValidation(IDataValidation):
         self.referenceTime = referenceTime
         
     def validate(self, series: Series) -> bool:
-        """ This method checks for missing date ranges in the the expected time series. 
+        """ This method checks for missing data points in the the expected time series. 
             :param series: Series - The series to validate
             :return: bool - True if the series passes validation, False otherwise
         """
-
-        if series.dataFrame is None or len(series.dataFrame) <= 0:
+        if series.dataFrame is None or series.dataFrame.empty:
             log_error('DateRangeValidation: No data in series to validate.')
             return False # No data to validate
     
@@ -64,6 +63,7 @@ class DateRangeValidation(IDataValidation):
             if time_difference > series.timeDescription.stalenessOffset:
                 log_error(f'DateRangeValidation: Series {series} is stale.')
                 log_error(f'Time difference: {time_difference}. Staleness offset: {series.timeDescription.stalenessOffset}')
+                log_error(f'Min timeGenerated: {series.dataFrame["timeGenerated"].min()}')
                 return False
         
         return True
--- a/src/ModelExecution/InputVectorBuilder.py
+++ b/src/ModelExecution/InputVectorBuilder.py
@@ -121,6 +121,15 @@ class InputVectorBuilder:
 
                     # Select only the wanted data
                     indexed_data = casted_data[index[0] : index[1]]
+
+                    # --- DEBUG ---
+                    if batchIndex == 0:
+                        timestamps = series.dataFrame['timeVerified'].tolist()
+                        sliced_ts = timestamps[index[0] : index[1]]
+                        print(f'\n[DEBUG] "{key}" sliced into vector (index [{index[0]}:{index[1]}]):')
+                        for ts, val in zip(sliced_ts, indexed_data):
+                            print(f'  {ts}  val={val}')
+                    # --- END DEBUG ---
                     
                     log(f'\t\t{key}: - amnt_found: {len(casted_data)}, indexed_len: {len(indexed_data)}')
                     # Concatenate the designated slice of casted data into the input vector
--- a/src/ModelExecution/InputVectorBuilder.py
+++ b/src/ModelExecution/InputVectorBuilder.py
@@ -121,15 +121,6 @@ class InputVectorBuilder:
 
                     # Select only the wanted data
                     indexed_data = casted_data[index[0] : index[1]]
-
-                    # --- DEBUG ---
-                    if batchIndex == 0:
-                        timestamps = series.dataFrame['timeVerified'].tolist()
-                        sliced_ts = timestamps[index[0] : index[1]]
-                        print(f'\n[DEBUG] "{key}" sliced into vector (index [{index[0]}:{index[1]}]):')
-                        for ts, val in zip(sliced_ts, indexed_data):
-                            print(f'  {ts}  val={val}')
-                    # --- END DEBUG ---
                     
                     log(f'\t\t{key}: - amnt_found: {len(casted_data)}, indexed_len: {len(indexed_data)}')
                     # Concatenate the designated slice of casted data into the input vector
--- a/src/ModelExecution/dataGatherer.py
+++ b/src/ModelExecution/dataGatherer.py
@@ -24,7 +24,7 @@ from DataValidation.IDataValidation import data_validation_factory
 from exceptions import Semaphore_Data_Exception, Semaphore_Ingestion_Exception
 from datetime import datetime, timedelta
 from pandas import date_range
-
+import copy
 
 
 class DataGatherer:
@@ -83,6 +83,8 @@ class DataGatherer:
 
         :param dependentSeriesList: list[DependentSeries] - The list of dependent series from the DSEPC
         :param referenceTime: datetime - The reference time to build the time description from.
+        :param key_to_index: dict[str, list[int]] - A mapping of outKey to vectorOrder index slice,
+            used to clip the series before validation so buffer slots are not validated.
 
         :returns: dict[str, Series] - The dictionary of the data it collected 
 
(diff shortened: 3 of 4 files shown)


src/ModelExecution/dataGatherer.py
```diff
@@ -119,10 +123,27 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
+                    series.timeDescription.interval,
+                    series.timeDescription.stalenessOffset
+                )
+                series_for_validation = Series.__new__(Series)
+                series_for_validation.__dict__.update(series.__dict__)
+                series_for_validation.dataFrame = trimmed_df
+                series_for_validation.timeDescription = trimmed_td
+                self.__validate_series(series_for_validation, referenceTime)
+            else:
+                self.__validate_series(series, referenceTime)
```
matdenkas: This needs to be put in a function

src/ModelExecution/dataGatherer.py
```diff
@@ -119,10 +123,27 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
+                    trimmed_from,
+                    trimmed_to,
+                    series.timeDescription.interval,
+                    series.timeDescription.stalenessOffset
+                )
+                series_for_validation = Series.__new__(Series)
+                series_for_validation.__dict__.update(series.__dict__)
+                series_for_validation.dataFrame = trimmed_df
+                series_for_validation.timeDescription = trimmed_td
+                self.__validate_series(series_for_validation, referenceTime)
```
matdenkas: Im uhhh not sure what you are trying to do here. It seems like you want a shallow or a deep copy? Either way there is a cleaner and easier way of doing this. (explicit copies are also more dangerous because if a new attribute is added your copy wont copy it)

```python3
import copy

# Shallow
series_for_validation = copy.copy(series)

# Deep
series_for_validation = copy.deepcopy(series)
```

I am not sure why you are copying here because validate_series does its own copy `src\DataValidation\DataValidationClasses\DateRangeValidation.py Line 36` If you need this copy for a specific reason you should put a comment maybe, its just a little hard for me to understand the why of what you are doing

src/ModelExecution/InputVectorBuilder.py
```diff
@@ -121,6 +121,15 @@ def __buildInputVector(self, vectorOrder: VectorOrder, dataRepository: dict[str,
 
                     # Select only the wanted data
                     indexed_data = casted_data[index[0] : index[1]]
+
+                    # --- DEBUG ---
+                    if batchIndex == 0:
+                        timestamps = series.dataFrame['timeVerified'].tolist()
+                        sliced_ts = timestamps[index[0] : index[1]]
+                        print(f'\n[DEBUG] "{key}" sliced into vector (index [{index[0]}:{index[1]}]):')
```
matdenkas: Should this be a print or a log?

matdenkas: Ah is this something you are meaning to have left in?

matdenkas (changes requested):

src/ModelExecution/InputVectorBuilder.py
lovelysandlonelys: an accidental print from before i got rid of my debugging prints

src/ModelExecution/dataGatherer.py
lovelysandlonelys: i am literally just dumb visa via silly way of copying, but the idea is that we only use the copy to validate but store the original series that's not clipped.

ccftissot:
> ## Current Behavior
> DataGatherer requests the input, pulling from the db or data ingestion. Next, data integrity processes are called (this is interpolation, we don't have any integrity classes that aren't interpolation). After interpolation we set the index to time verified and reindex the series to the interval used in the model. Finally the data is validated and then stored in the series_repository where it will later go through post processing and vector ordering before being passed to the model.
> 
> The point of contention here is validation, we need to ensure only the data necessary for the model to run is being validated.

First of, thank you for all your work on this. I think this is crucial to get right for the future of Semaphore. And thank you for documenting the process and making it way clearer.  My take on this is that I am not sure we are doing all this in the correct order.  Here's a diagram (my mind works better with diagrams)  of what I think the current process is based on your description:

image


**Conceptually** i think the below process makes more sense:

image

Here's my thinking on this:  we want the post processing to get as much data as possible based on the dspec (because we have no idea what each model may want to do with that data).  So we should give it the full range (the requested range) and we should give it after we have interpolated to try and fix any possible gaps. 

However, I don't think we should validate the data before passing it to the post processor. Who knows what this guy is going to do?  some of it drops a bunch of data by calculating moving averages and other such bizarries. 

After all that is done, we should worry about formatting the final vector for input to the model. This is where we drop indexes if needed based on the final input range as specified in the vertorOrder property.   Finally when all this is done, we can check to see if we are missing any data and if we are then we need to fail.

I know this is possibly a big change compared to what we do today, so let's discuss at our next meeting. Some things to consider:  do any or our post processing classes rely on validation having happened first?  what other data validation do we do besides mapping from the requested range to the final input (vectorOrder) range? and do these require more data than the model range?

I am going to block the review until we have a chance to discuss.

ccftissot (changes requested):
Have not looked at the code yet but blocking the review until we get a chance to discuss the overall approach.

src/DataValidation/DataValidationClasses/DateRangeValidation.py
```diff
@@ -25,12 +25,11 @@ def __init__(self, referenceTime: datetime = None):
         self.referenceTime = referenceTime
         
     def validate(self, series: Series) -> bool:
-        """ This method checks for missing date ranges in the the expected time series. 
+        """ This method checks for missing data points in the the expected time series.
```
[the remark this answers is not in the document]
lovelysandlonelys: done

src/DataValidation/DataValidationClasses/DateRangeValidation.py
```diff
@@ -64,6 +63,7 @@ def validate(self, series: Series) -> bool:
             if time_difference > series.timeDescription.stalenessOffset:
                 log_error(f'DateRangeValidation: Series {series} is stale.')
                 log_error(f'Time difference: {time_difference}. Staleness offset: {series.timeDescription.stalenessOffset}')
+                log_error(f'Min timeGenerated: {series.dataFrame["timeGenerated"].min()}')
```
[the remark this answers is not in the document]
lovelysandlonelys: done

src/ModelExecution/dataGatherer.py
```diff
@@ -56,16 +56,20 @@ def get_data_repository(self, dspec: Dspec, referenceTime: datetime) -> dict[str
         dependentSeries: list[DependentSeries] = dspec.dependentSeries
         postProcessCalls: list[PostProcessCall] = dspec.postProcessCall
 
+        # Build a dict of {key: index} so validation knows the actual consumed window
+        vector_order = dspec.orderedVector
+        key_to_index = dict(zip(vector_order.keys, vector_order.indexes))
+        
         # Get Dependent Data
-        dependent_data_repository = self.__request_dependent_data(dependentSeries, referenceTime)
+        dependent_data_repository = self.__request_dependent_data(dependentSeries, referenceTime, key_to_index)
```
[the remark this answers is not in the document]
lovelysandlonelys: this shouldn't be an issue in the meantime because if we over request we would be doing so with interpolation in mind so thee should always be an outkey. we do want to move validation after post processing but that is a refactor for a later date

src/ModelExecution/dataGatherer.py
```diff
@@ -119,15 +125,64 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
+            # Build a corrected TimeDescription whose fromDateTime and toDateTime
+            # match the actual first and last rows of the trimmed slice. DateRangeValidation
+            # uses these to construct its expected index — if they still pointed at the
+            # full over-requested window, validation would flag the trimmed rows as missing.
+            trimmed_td = TimeDescription(
+                trimmed_df['timeVerified'].iloc[0],
+                trimmed_df['timeVerified'].iloc[-1],
+                series.timeDescription.interval,
+                series.timeDescription.stalenessOffset
+            )
```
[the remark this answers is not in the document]
lovelysandlonelys: this seems like overkill to me

lovelysandlonelys:
There will be a new PR with the fixes for this branch alongside changes discussed in the meeting (slicing correctly and passing clipped series only).

PR #957, opened 2026-04-23
ADDITIONAL HOUR INTO FUTURE

### LINKED ISSUE #920: Overrequest Lighthouse Minute Data

lovelysandlonelys:
Requesting an hour of lighthouse data into the future, merging into the branch that this fix depends on. If you test right now it will say that there are still nans in the series when it reaches validation. Just want to separate concerns and have people double check that all relevant lighthouse calls are extended into the future. Further testing to do once merged but I ran a few on the other branch and seems to work as expected. The other ticket regarding indexes should stay as 2, 27 because this point would fall at index 27 and the slice is exclusive. 

image

### DIFF
--- a/data/dspec/CRPS/CRPS_108hr.json
+++ b/data/dspec/CRPS/CRPS_108hr.json
@@ -30,7 +30,7 @@
             "unit": "celsius",
             "interval": 3600,
             "range": [
-                0,
+                1,
                 -26
             ],
             "outKey": "south-bird-island_water-temp_25",
@@ -51,7 +51,7 @@
             "unit": "celsius",
             "interval": 3600,
             "range": [
-                0,
+                1,
                 -26
             ],
             "outKey": "south-bird-island_air-temp_25",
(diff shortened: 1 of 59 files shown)


ccftissot (changes requested):
not sure what is going @lovelysandlonelys but i am having a lot of problems with this branch.  Did you rename it? even after resetting my local branch to point to the (new? renamed?) branch,  I do not see any of the changes that the PR is showing.  can you please provide a description of each branch and what changes are expected in them?

ccftissot:
> not sure what is going @lovelysandlonelys but i am having a lot of problems with this branch. Did you rename it? even after resetting my local branch to point to the (new? renamed?) branch, I do not see any of the changes that the PR is showing. can you please provide a description of each branch and what changes are expected in them?

ok, so it's the 920 branch from the 919 branch .  i got the code now

ccftissot (changes requested):
Looks like we are missing some changes? or is this intentional? 

ThermalRefuge:  ranges have not been changed for water temp and air temp measurements.

Testmodels also have not been changed except for the CRPS 120 but I am assuming that is on purpose/doesn't matter

lovelysandlonelys:
combined to one pr

Issue #930, opened 2026-04-06
Thermal Refuge Performance Discrepencies 


lovelysandlonelys:
In an [analysis of performance](https://docs.google.com/document/d/1rn-II8nBFOUU6a1ycLZMjlW5HPhUkjSc/edit?usp=sharing&ouid=101272362728836938247&rtpof=true&sd=true) on dev and prod it was found that Thermal Refuge performs much better on dev despite the despcs being exactly the same. We're theorizing that maybe dev runs into more enginex issues?

PR #961, opened 2026-04-29
919 what if we sliced in a normal way

### LINKED ISSUE #919: Lighthouse Haunted Data Gap Fix

### LINKED ISSUE #920: Overrequest Lighthouse Minute Data

### LINKED ISSUE #952: Indexing Bug
Currently for lighthouse models we over request data in the past in order to ensure room for interpolatin. There are two tickets in the works to make this process more useful/sustainable 1) to see if we can also over request into the futrue, and 2) clipping the date range so that we only validate on what the model actually needs, not the entire over requested range. 

This error was found while working on these two tickets. In main right now this is the current behavior: 

```
[DEBUG] Series "south-bird-island_air-temp_25" post-validation:
  Reference time: 2026-04-21 16:00:00+00:00
  Shape: (27, 7)
  Expected range: 2026-04-20 14:00:00+00:00 ΓåÆ 2026-04-21 16:00:00+00:00  interval=1:00:00
  Null count: 0 / 27
                timeVerified    offset_from_ref dataValue
0  2026-04-20 14:00:00+00:00  -2 days +22:00:00      21.1
1  2026-04-20 15:00:00+00:00  -2 days +23:00:00      19.5
2  2026-04-20 16:00:00+00:00  -1 days +00:00:00      20.5
3  2026-04-20 17:00:00+00:00  -1 days +01:00:00      20.2
4  2026-04-20 18:00:00+00:00  -1 days +02:00:00      20.2
..                       ...                ...       ...
22 2026-04-21 12:00:00+00:00  -1 days +20:00:00      23.0
23 2026-04-21 13:00:00+00:00  -1 days +21:00:00      22.9
24 2026-04-21 14:00:00+00:00  -1 days +22:00:00      23.6
25 2026-04-21 15:00:00+00:00  -1 days +23:00:00      23.8
26 2026-04-21 16:00:00+00:00    0 days 00:00:00      24.3
```

This is the data post-validation (has been interpolated and we have the entire date range). 

This is the data going into the model after indexing: 
```
[DEBUG] "south-bird-island_air-temp_25" sliced into vector (index [0:25]):
  2026-04-20 14:00:00+00:00  val=21.1
  2026-04-20 15:00:00+00:00  val=19.5
  2026-04-20 16:00:00+00:00  val=20.5
  2026-04-20 17:00:00+00:00  val=20.2
  2026-04-20 18:00:00+00:00  val=20.2
  2026-04-20 19:00:00+00:00  val=20.4
  2026-04-20 20:00:00+00:00  val=20.6
  2026-04-20 21:00:00+00:00  val=21.0
  2026-04-20 22:00:00+00:00  val=21.2
  2026-04-20 23:00:00+00:00  val=20.9
  2026-04-21 00:00:00+00:00  val=22.3
  2026-04-21 01:00:00+00:00  val=21.9
  2026-04-21 02:00:00+00:00  val=22.7
  2026-04-21 03:00:00+00:00  val=22.7
  2026-04-21 04:00:00+00:00  val=22.7
  2026-04-21 05:00:00+00:00  val=22.9
  2026-04-21 06:00:00+00:00  val=22.9
  2026-04-21 07:00:00+00:00  val=22.8
  2026-04-21 08:00:00+00:00  val=22.8
  2026-04-21 09:00:00+00:00  val=23.0
  2026-04-21 10:00:00+00:00  val=22.4
  2026-04-21 11:00:00+00:00  val=22.7
  2026-04-21 12:00:00+00:00  val=23.0
  2026-04-21 13:00:00+00:00  val=22.9
  2026-04-21 14:00:00+00:00  val=23.6
04/21/26 16:25:05: 		south-bird-island_air-temp_25: - amnt_found: 27, indexed_len: 25
```

The issue here is that rather than getting ride of the over requested values in the past, our current indexing on the series [0, 25] clips off the now time and the last our of measurements. 

This is not the behavior we want as the models should be ingesting the past 24 hours of air temperature measurements and the now point. I thought it would be better to make this a separate bug fix issue before tackling the other two tickets. 

This is critical as this means that as of this moment, both semaphore dev and prod are going feeding the cold stunning models the wrong data.

lovelysandlonelys:
### Summary
Closes #919.

This PR bundles three related fixes that all stem from the same root cause: DateRangeValidation was running against the full over-requested data window, including interpolation buffer slots that the model never actually consumes. A single Lighthouse dropout would produce buffer-slot nulls that looked like missing required data, causing two consecutive failure batches even though the model had everything it needed.

1. Clip series to the vectorOrder window before validation (DataGatherer)

DataGatherer.__request_dependent_data now calls __clip_and_validate_series, which slices the series down to only the rows specified by the vectorOrder index (iloc[start:end+1]) before passing it to DateRangeValidation. It also rebuilds the TimeDescription to match the trimmed window so the validator's expected date range aligns correctly. The full series (including buffer slots) is still fetched and stored in the repository so downstream vectorOrder indexing works as before.

2. Make vectorOrder indexing inclusive and extend Lighthouse ranges by one future hour

The vectorOrder index slice was previously exclusive on the upper bound, meaning [0, 25] only consumed 25 rows (0–24) rather than 26. Indexes across all dspecs have been corrected to be inclusive (iloc[start:end+1]). Simultaneously, all Lighthouse dependentSeries that use hourly interpolation had their range toOffset bumped from 0 to 1 — requesting one additional future hour so interpolation has a bracket on both sides of the current timestamp when the exact "now" reading is missing.

3. dspec updates across all models (ColdStunning, CRPS, MRE, Bird Island, Magnolia, MLP-OP)

Every dspec that pulls from Lighthouse with an hourly interval received:

range toOffset: 0 → 1 (one extra future hour for interpolation)
vectorOrder indexes corrected to the new inclusive convention (start shifted from 0 to 2 for Lighthouse inputs to skip the buffer slot; end values adjusted to be inclusive)
dspec version bumped to 2.1

### Testing

1. Run tests: `docker exec semaphore-core python3 -m pytest src/tests/`
2. Test some models: 
    - `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/MRE_Bird-Island_Water-Temperature_12hr.json -v`
    -  `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Magnolia/12/magnolia_12.json -v`
         - Magnolia has been failing for me with all data, not sure what that's about I need to double check if this is a lighthouse thing or a semaphore thing 💀  
    - `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_12hr.json -v`
    - `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/MLP-OP.json -v`
3. Double check my work ;-;


src/ModelExecution/dataGatherer.py
```diff
@@ -119,15 +124,65 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
+
+            series.dataFrame = trimmed_df
+            series.timeDescription = trimmed_td
+
+            log(f'[DataGatherer] Clipped series for key "{key}" @ {referenceTime}:\n'
+                f'\t  rows      : {len(series.dataFrame)}\n'
+                f'\t  first     : {series.dataFrame["timeVerified"].iloc[0]}\n'
+                f'\t  last      : {series.dataFrame["timeVerified"].iloc[-1]}\n'
+                f'\t  td.from   : {series.timeDescription.fromDateTime}\n'
+                f'\t  td.to     : {series.timeDescription.toDateTime}')
```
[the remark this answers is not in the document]
lovelysandlonelys: these will not be log unless the verbose flag is on

src/orchestrator.py
```diff
@@ -78,6 +78,17 @@ def run_semaphore(self, dspecPaths: list[str], executionTime: datetime = None, t
...
+                    
+                    # I want a print statment here of what the input vectors look like before they go into the model runner. 
+                    # This is a critical point that will help with debugging and ensuring that the data is being gathered and processed correctly.
+                    lines = [f'[Orchestrator] Input vectors for {model_name} @ {reference_time}']
+                    for k in DSPEC.orderedVector.keys:
+                        df = data_repository[k].dataFrame
+                        lines.append(f'\t[{k}]')
+                        for i, (_, row) in enumerate(df.iterrows()):
+                            lines.append(f'\t\t{i+1}. {row["timeVerified"]} : {row["dataValue"]}')
+                    log('\n'.join(lines))
```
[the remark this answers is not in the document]
lovelysandlonelys: this will not print unless the -v flag is on the run

src/ModelExecution/dataGatherer.py
```diff
@@ -119,15 +124,65 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
-            self.__validate_series(series, referenceTime)
+            # Validate only the rows the model will actually consume, not the full
+            # over-requested window including interpolation buffer slots.
+            # The full series (all rows) still goes into the repository so vectorOrder
+            # can index it normally downstream.
+            series = self.__clip_and_validate_series(series, key, key_to_index, referenceTime)
             
-            # Store the series in the repository
+            # Clipped Series (only points that the model actually wantes) goes in the repo
             series_repository[key] = series
```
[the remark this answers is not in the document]
lovelysandlonelys: removed inconsistant comment

src/ModelExecution/InputVectorBuilder.py
```diff
@@ -91,21 +91,23 @@ def __buildInputVector(self, vectorOrder: VectorOrder, dataRepository: dict[str,
...
+
                 # Grab all data, this changes if its a multi series or not
                 data = None
                 isFinished = True # Assume we are finished unless we find a multi series that has more data
                 if not keyIsMulti:
                     data = series.dataFrame['dataValue'].to_list()
+                    if index[0] is not None:
+                        data = data[index[0]:index[1]]
                 else:
                     data = [ensemble_members[batchIndex] for ensemble_members in series.dataFrame['dataValue']] # We pull a slice of data determined by the batch index
```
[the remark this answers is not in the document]
lovelysandlonelys: this is a bug that leaked through when i was trying to fix the vector order tests, has since been removed so we are not clipping twice

src/orchestrator.py
```diff
@@ -78,6 +78,17 @@ def run_semaphore(self, dspecPaths: list[str], executionTime: datetime = None, t
 
                     data_repository = self.dataGatherer.get_data_repository(DSPEC, reference_time)
                     input_vectors = self.inputVectorBuilder.build_batch(DSPEC, data_repository)
+                    
+                    # I want a print statment here of what the input vectors look like before they go into the model runner. 
+                    # This is a critical point that will help with debugging and ensuring that the data is being gathered and processed correctly.
```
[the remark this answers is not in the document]
lovelysandlonelys: fixed

src/ModelExecution/dataGatherer.py
```diff
@@ -56,16 +55,20 @@ def get_data_repository(self, dspec: Dspec, referenceTime: datetime) -> dict[str
         dependentSeries: list[DependentSeries] = dspec.dependentSeries
         postProcessCalls: list[PostProcessCall] = dspec.postProcessCall
 
+        # Build a dict of {key: index} so validation knows the actual consumed window
+        vector_order = dspec.orderedVector
+        key_to_index = dict(zip(vector_order.keys, vector_order.indexes))
+
```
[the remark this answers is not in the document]
lovelysandlonelys: Every vectorOrder key maps 1:1 to a unique outKey from dependentSeries — a key appearing twice has never been a pattern in any of the dspecs and isn't part of the design.

src/ModelExecution/dataGatherer.py
```diff
@@ -119,15 +124,65 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
+            # Build a corrected TimeDescription whose fromDateTime and toDateTime
+            # match the actual first and last rows of the trimmed slice. DateRangeValidation
+            # uses these to construct its expected index — if they still pointed at the
+            # full over-requested window, validation would flag the trimmed rows as missing.
+            trimmed_td = TimeDescription(
+                trimmed_df['timeVerified'].iloc[0],
+                trimmed_df['timeVerified'].iloc[-1],
+                series.timeDescription.interval,
+                series.timeDescription.stalenessOffset
+            )
```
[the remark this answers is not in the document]
lovelysandlonelys: The indexes come directly from a dspec that the team authors — if index[1] is None or out of range, the dspec is malformed and I assume we want it to blow up loudly with an unhandled exception. The existing code has no defensive guards on dspec values anywhere else (range, interval, leadTime, etc.) and that's intentional — bad dspec = crash = fix the dspec.

data/dspec/ColdStunning/Bird-Island_Water-Temperature_18hr.json
```diff
@@ -84,29 +84,8 @@
     ],
     "postProcessCall": [],
     "vectorOrder": [
-        {
```
abeasley1722, line 87: Great work!

data/dspec/ColdStunning/Bird-Island_Water-Temperature_36hr.json
```diff
@@ -84,29 +84,8 @@
...
-            "key": "south-bird-island_predicted_air-temp_42",
-            "dType": "float",
-            "indexes": [
-                0,
-                36
-            ]
-        }
+        { "key": "south-bird-island_water-temp_25"        , "dType": "float", "indexes": [2, 26] },
+        { "key": "south-bird-island_air-temp_25"          , "dType": "float", "indexes": [2, 26] },
+        { "key": "south-bird-island_predicted_air-temp_42", "dType": "float", "indexes": [0, 35] }
```
abeasley1722, line 89: Is the last one indexed differently because it is predictions?

data/dspec/ColdStunning/Bird-Island_Water-Temperature_18hr.json
abeasley1722: Shifting the index to not include the buffer values was a good idea

data/dspec/ColdStunning/MRE_Bird-Island_Water-Temperature_114hr.json
```diff
@@ -89,30 +89,8 @@
...
-            "key": "south-bird-island_air-temp_25",
-            "dType": "float",
-            "indexes": [
-                0,
-                25
-            ]
-        },
-        {
-            "key": "ensemble-south-bird-island_predicted_air-temp_114",
-            "dType": "float",
```
abeasley1722, line 110: I thought that if we don't give the right amount of data the model breaks. So, I wonder why it was fine with us give more data than needed?

src/DataValidation/DataValidationClasses/DateRangeValidation.py
```diff
@@ -25,12 +25,11 @@ def __init__(self, referenceTime: datetime = None):
         self.referenceTime = referenceTime
         
     def validate(self, series: Series) -> bool:
-        """ This method checks for missing date ranges in the the expected time series. 
+        """ This method checks for missing data points in the expected time series.
```
abeasley1722, line 28: Good catch

abeasley1722 (dismissed):
Works as intended. Good job!

data/dspec/ColdStunning/Bird-Island_Water-Temperature_36hr.json
lovelysandlonelys: got it in one, yes

data/dspec/ColdStunning/Bird-Island_Water-Temperature_18hr.json
lovelysandlonelys: yippeee

data/dspec/ColdStunning/MRE_Bird-Island_Water-Temperature_114hr.json
lovelysandlonelys: It would break if we actually handed the model the larger data vector, but because we index it's fine.

src/tests/UnitTests/test_inputVectorBuilder.py
```diff
@@ -132,36 +132,41 @@ def test_build_batch_single_value_input():
...
-    inputVectorBuilder = InputVectorBuilder()
-    assert inputVectorBuilder._InputVectorBuilder__cast_value('1.0', 'float') == 1.0
+#     # Compare expected vs Actual
+#     assert len(EXPECTED_RESULT) == len(result)
+#     for expected, actual in zip(EXPECTED_RESULT, result):
+#         assert expected == actual
+
+# def test__cast_value():
+#     inputVectorBuilder = InputVectorBuilder()
+#     assert inputVectorBuilder._InputVectorBuilder__cast_value('1.0', 'float') == 1.0
```
matdenkas: We should never be leaving dead code

src/ModelExecution/dataGatherer.py
```diff
@@ -119,15 +124,63 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
+                f'\t  td.from   : {series.timeDescription.fromDateTime}\n'
+                f'\t  td.to     : {series.timeDescription.toDateTime}')
+
+        self.__validate_series(series, referenceTime)
+        return series
+
+    
     def __post_process_data(self, series_repository: dict[str, Series], postProcessCalls: list[PostProcessCall]) -> dict[str, Series]:
         """
         This function calls the post processing methods for any inputs that need post processing.
```
matdenkas: Not blocking but functions really shouldnt be doing more than one thing. If you have to put and in a function name imo thats a call that that function needs to be split

src/orchestrator.py
```diff
@@ -78,6 +78,17 @@ def run_semaphore(self, dspecPaths: list[str], executionTime: datetime = None, t
...
+                    
+                    # I want a print statement here of what the input vectors look like before they go into the model runner. 
+                    # This is a critical point that will help with debugging and ensuring that the data is being gathered and processed correctly.
+                    lines = [f'[Orchestrator] Input vectors for {model_name} @ {reference_time}']
+                    for k in DSPEC.orderedVector.keys:
+                        df = data_repository[k].dataFrame
+                        lines.append(f'\t[{k}]')
+                        for i, (_, row) in enumerate(df.iterrows()):
+                            lines.append(f'\t\t{i+1}. {row["timeVerified"]} : {row["dataValue"]}')
+                    log('\n'.join(lines))
```
matdenkas: Super not blocking but this should be encapsulated. This is absolutely nitpicky but the reason is `run_semaphore` is a Controller or Delegator function, it in its self is not really making any decisions its just runny an algorithm. Logging in itself wouldnt be wrong, but having loops and string manipulation over several lines in the middle of the function looks inherently less clean to me. Absolutely not blocking and absolutely my opinion

matdenkas (dismissed):
I wrote some comments about code structure, but they are not blocking . Tests pass, code works nicely done!

src/ModelExecution/dataGatherer.py
lovelysandlonelys: fixed

src/orchestrator.py
lovelysandlonelys: put into function

matdenkas (dismissed):
Tests passed, code seems cleaner to me. Thanks!

CJQuintero61 (dismissed):
Good from me, just resolve the conflicts. I had to draw things out to understand what was going on, but things then made sense.

lovelysandlonelys:
Ran branch on `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_6hr.json`, `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/VirginiaKey/12hr_VirginiaKey_wl_Estrada.json `, and `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Surge/Aransas/ar_mlp_surge_12h_pred.json ` all of which ran sucesfully and all of which call the resolve vector components post processing class, ar_mlp also calls some arithmatic operations that worked. I tried to run 'docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Inundation/April/ar_inundation_apr_12h.json' but still got the same NDBC error we discussed in the monday meeting so the 4 max mean post processing class lives in mystery, but other classes appear to be working. 

I resolved a merge conflict and just need the people who reviewed before to re-review 😭

data/dspec/ThermalRefuge.json
ccftissot, line 46: should we increase this guy too?

ccftissot, line 28: should we increase this guy too?

data/dspec/MLP-OP.json
```diff
@@ -35,7 +35,7 @@
             "series": "dAirTmp",
             "unit": "celsius",
             "interval": 3600,
-            "range": [0, -23],
+            "range": [1, -23],
```
ccftissot: should we request more and trim the results as in all other models?

data/dspec/MLP-OP.json
```diff
@@ -25,7 +25,7 @@
             "series": "dWaterTmp",
             "unit": "celsius",
             "interval": 3600,
-            "range": [0, -23],
+            "range": [1, -23],
```
ccftissot: should we request more and trim the results as in all other models? i.e., [1, -25]

src/ModelExecution/dataGatherer.py
```diff
@@ -119,15 +124,64 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
             self.__validate_series(series, referenceTime)
             
-            # Store the series in the repository
+            # Clipped Series (only points that the model actually wants) goes in the repo
             series_repository[key] = series
 
         return series_repository
 
 
+    def __clip_series(self, series: Series, key: str, key_to_index: dict[str, list[int]], referenceTime: datetime):
```
ccftissot: why pass the whole key_to_index dictionary here if this method only looks at one dependent series at the time? Why not just pass the index directly? always try to  minimize the arguments needed by a method.

ccftissot: we definitely need to define a unit test for this method.

src/ModelExecution/dataGatherer.py
```diff
@@ -119,15 +124,64 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
+                trimmed_df['timeVerified'].iloc[0],
+                trimmed_df['timeVerified'].iloc[-1],
+                series.timeDescription.interval,
+                series.timeDescription.stalenessOffset
+            )
+
+            series.dataFrame = trimmed_df
+            series.timeDescription = trimmed_td
+
+            log(f'[DataGatherer] Clipped series for key "{key}" @ {referenceTime}:\n'
```
ccftissot, line 174: will this cause it to log all the time? should we only log at information level? or skip logging altogether unless there is an exception?

src/orchestrator.py
```diff
@@ -82,6 +82,9 @@ def run_semaphore(self, dspecPaths: list[str], executionTime: datetime = None, t
 
                     data_repository = self.dataGatherer.get_data_repository(DSPEC, reference_time)
                     input_vectors = self.inputVectorBuilder.build_batch(DSPEC, data_repository)
+                    
+                    self._log_input_vectors(model_name, reference_time, data_repository, DSPEC)
```
ccftissot, line 86: will this cause it to always log?  or only when verbose is turned on?  I don't think we want to always log this.

ccftissot (commented):
just a few comments.  i did not test this but i believe others did.  let me know if we should discuss.

data/dspec/ThermalRefuge.json
lovelysandlonelys: I personally don't think it's necessary as we've never had issues with the first point for thermal refuge in comparison to the other models but if you'd like to do it to be safe let me know. Thermal refuge just has a lot more indexing so I didn't want to complicate that if I didn't have to.

src/ModelExecution/dataGatherer.py
lovelysandlonelys: We only log the things that start with just the log tag if the verbose flag is turned on. Otherwise we only log things that are tagged with the failure or success tag I believe. I figured it was fine to leave here since it is a tool for debugging when running with the verbose flag but please let me know if you think I should remove this, I think it's useful for testing.

src/orchestrator.py
lovelysandlonelys: Only with the verbose flag! We won't always log this.

data/dspec/ThermalRefuge.json
lovelysandlonelys: Same reasoning as above but again let me know if you think this is an over-site. I really think it'll be simpler to leave thermal refuge along as much as possible though since we already have active issues with thermal refuge: https://github.com/conrad-blucher-institute/semaphore/issues/930 &  https://github.com/orgs/conrad-blucher-institute/projects/17/views/8?pane=issue&itemId=79651409&issue=conrad-blucher-institute%7Csemaphore%7C434

src/ModelExecution/dataGatherer.py
lovelysandlonelys: There's a test in src/tests/UnitTests/test_dataGatherer.py that checks that the series is clipped before going to validation. Is that not equivalent to a unit test for the method? Should I add another test just for this function, if so where did I go wrong with the other test 😅

data/dspec/MLP-OP.json
lovelysandlonelys: yes, i think i missed this one on my first go through ages ago and forgot to extend on the other side

data/dspec/MLP-OP.json
lovelysandlonelys: yes, i think i missed this one on my first go through ages ago and forgot to extend on the other side

lovelysandlonelys: I corrected this however when you test it will fail because currently Packery is one of the Lighthouse stations not giving data: 

https://lighthouse.tamucc.edu/pd?stnlist=005&serlist=atp&when=05.10.2026-05.13.2026&whentz=UTC0&-action=g&unit=metric&elev=stnd&style=line

However I was able to confirm that the clip looks right: 
image

src/ModelExecution/dataGatherer.py
lovelysandlonelys: fixed! just lookup the key in the dictionary and pass that.

abeasley1722 (approved):
Works as intended!

data/dspec/Magnolia/12/magnolia_12.json
```diff
@@ -82,7 +82,7 @@
             "series": "dWnDir",
             "unit": "degrees",
             "interval": 3600,
-            "range": [0, -26],
+            "range": [1, -26],
```
CJQuintero61, line 85: Would this make the magnolia models always fail since we are requesting a measurement into the future?

CJQuintero61 (approved):
I'm going to approve so that we can push this to dev and continue monitoring the changes there, but I do have a comment I would like explained sometime.

Wouldn't requesting an extra hour for magnolia models always fail since we are requesting a measurement 1 hour into the future?

data/dspec/Magnolia/12/magnolia_12.json
CJQuintero61: resolving to allow the merge

PR #972, opened 2026-06-02
968 magnolia series not clipping

### LINKED ISSUE #968: Magnolia Series Not Clipping
The series goes through post processing and somehow hits date range validation before we clip the series.

lovelysandlonelys:
### Problem

Dependent series that are consumed exclusively by post-process calls (i.e. not listed directly in `vectorOrder`) were failing `DateRangeValidation` with false negatives.

The existing clip logic in `__clip_series` correctly trims a series to only the rows the model actually consumes before validation, using `key_to_index` (built from `vectorOrder`). However, if a series' `outKey` does not appear in `vectorOrder` — because it feeds a post-process call that produces a *different* output key — `key_to_index.get(key)` returns `None` and the clip is skipped entirely. The full over-requested window, including interpolation buffer slots, is then passed to `DateRangeValidation`.

This caused models to fail at runtime when the leading buffer slot (e.g. `T+0` or the current hour) had not yet been ingested, even though that slot is never read by the model.

**Example:** In `magnolia_12`, `PL_WNDIR_25` and `PL_WNSPD_25` both have `range: [1, -26]` and feed `ResolveVectorComponents`, which outputs `x_PL_PWNDCMP_25` / `y_PL_PWNDCMP_25`. Those derived keys appear in `vectorOrder` with `indexes: [2, 26]`, but the raw wind keys do not. At runtime (`21:59 UTC`), `toDateTime` resolved to `22:00 UTC` — a slot that didn't exist yet — causing validation to fail on wind direction while all other series passed cleanly.

### Fix

Added `__resolve_indexes` to `DataGatherer`. When a key is not found directly in `key_to_index`, the method walks `postProcessCalls` looking for a call whose args contain the key as an `_inKey` value, then returns the `key_to_index` entry for the corresponding `_outKey`. This traces the key one step through post-processing to find the indexes of the vectorOrder entry it ultimately feeds.

This relies on the existing convention that post-process args use `_inKey` / `_outKey` suffixes for their input and output key references, which is consistent across all current post-process calls.

`__request_dependent_data` was updated to call `__resolve_indexes` instead of `key_to_index.get(key)` directly, and `postProcessCalls` is now passed through from `get_data_repository`.

This is technically another work around for the [restructure](https://github.com/conrad-blucher-institute/semaphore/issues/919#issuecomment-4359064516) that we spoke about the first time this issue came up, however I maintain the stance that a refactor would be more time consuming than simply solving this bug in a slightly silly way. 

### Files Changed

- `src/ModelExecution/dataGatherer.py`
  - Added `__resolve_indexes`
  - Updated `__request_dependent_data` signature to accept `postProcessCalls`
  - Updated clip call to use `__resolve_indexes`
  - Updated call site in `get_data_repository`

### Testing

1. `docker compose up -d --build`
2. `docker exec semaphore-core python3 -m pytest src/tests/ --run-slow `
3. `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/Magnolia/48/magnolia_48.json -v`
4. `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/MRE_Bird-Island_Water-Temperature_102hr.json`
5. `docker exec semaphore-core python3 src/semaphoreRunner.py -d ./data/dspec/ColdStunning/Bird-Island_Water-Temperature_102hr.json`

### DIFF
--- a/data/dspec/Inundation/April/ar_inundation_apr_12h.json
+++ b/data/dspec/Inundation/April/ar_inundation_apr_12h.json
@@ -68,7 +68,7 @@
             "args": {
                 "target_inKey": "DPD_12",
                 "warning_trigger": "20",
-                "outkey": "fmm_DPD_1" 
+                "outKey": "fmm_DPD_1" 
                      
             }
         },
@@ -77,7 +77,7 @@
             "args": {
                 "target_inKey": "APD_12",
                 "warning_trigger": "20",
-                "outkey": "fmm_APD_1" 
+                "outKey": "fmm_APD_1" 
                      
             }
         },
@@ -86,7 +86,7 @@
             "args": {
                 "target_inKey": "WVHT_12",
                 "warning_trigger": "20",
-                "outkey": "fmm_WVHT_1" 
+                "outKey": "fmm_WVHT_1" 
                      
             }
         },
@@ -95,7 +95,7 @@
             "args": {
                 "target_inKey": "dWl_12",
                 "warning_trigger": "11",
-                "outkey": "fmm_dWl_1" 
+                "outKey": "fmm_dWl_1" 
                      
             }
         }
--- a/src/ModelExecution/dataGatherer.py
+++ b/src/ModelExecution/dataGatherer.py
@@ -155,8 +155,8 @@ def __resolve_indexes(self, key: str, key_to_index: dict, postProcessCalls: list
         # Key is not in vectorOrder — search post-process calls for one that
         # consumes this key as an inKey, then return the indexes of its outKey
         for call in postProcessCalls:
-            in_keys  = [v for k, v in call.args.items() if k.endswith('_inKey')]
-            out_keys = [v for k, v in call.args.items() if k.endswith('_outKey')]
+            in_keys  = [v for k, v in call.args.items() if k.lower().endswith('_inkey')]
+            out_keys = [v for k, v in call.args.items() if k.lower().endswith('_outkey')]
             if key in in_keys:
                 for out_key in out_keys:
                     if out_key in key_to_index:
--- a/src/ModelExecution/dataGatherer.py
+++ b/src/ModelExecution/dataGatherer.py
@@ -68,7 +68,7 @@ class DataGatherer:
         return post_processed_series_repository
     
 
-def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], referenceTime: datetime, key_to_index: dict[str, list[int]], postProcessCalls: list[PostProcessCall]) -> dict[str, Series]:
+    def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], referenceTime: datetime, key_to_index: dict[str, list[int]], postProcessCalls: list[PostProcessCall]) -> dict[str, Series]:
         """This method handles the process of requesting the dependant series from the DSPEC. Its requests will be temporally
         referenced from the passed reference time. It will:
             - Build the series description
@@ -135,7 +135,7 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
         return series_repository
 
 
-def __resolve_indexes(self, key: str, key_to_index: dict, postProcessCalls: list[PostProcessCall]) -> tuple | None:
+    def __resolve_indexes(self, key: str, key_to_index: dict, postProcessCalls: list[PostProcessCall]) -> tuple | None:
         """Resolves the vectorOrder indexes for a given key. If the key is directly in
         vectorOrder, returns its indexes. If the key is only consumed by a post-process
         call (i.e. not in vectorOrder directly), walks the post-process args to find the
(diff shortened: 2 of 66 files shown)


src/ModelExecution/dataGatherer.py
```diff
@@ -136,6 +135,35 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
+        :param postProcessCalls: list[PostProcessCall] - The post-process calls from the dspec
+        :returns: tuple | None - The (start, end) index tuple, or None if unresolvable
+        """
+        # Direct hit — key is in vectorOrder
+        if key in key_to_index:
+            return key_to_index[key]
+
+        # Key is not in vectorOrder — search post-process calls for one that
+        # consumes this key as an inKey, then return the indexes of its outKey
+        for call in postProcessCalls:
```
ccftissot, line 168: i think that this will work for now. It does restrict the chain of keys to a 2-level chain.  If we had, say, 

s1, s2 not in vector order
s1 fed to postprocess p1 that generate s1b
postprocess p2 uses s1b and s2 to produce s3
only s3 is in vector order.

this code would not work.  I don't think we should worry about fixing that for now, but let's put a comment/TODO in here that says it doesn't support this scenario.

src/ModelExecution/dataGatherer.py
```diff
@@ -126,8 +126,7 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
 
             # Validate only the rows the model will actually consume, not the full
             # over-requested window including interpolation buffer slots.
-            series = self.__clip_series(series, key, key_to_index.get(key), referenceTime)
-
+            series = self.__clip_series(series, key, self.__resolve_indexes(key, key_to_index, postProcessCalls), referenceTime)
             self.__validate_series(series, referenceTime)
```
ccftissot, line 136: if __resolve_indexes return None, will vaildate_series error out?  what should we do if the series is not in vectororder and not used by a postprocess? should we just ignore the series at this point?

ccftissot (dismissed):
I think that this is actually a nice temporary solution to the problem. it does add to the argument that our ingestion process needs to be refactored; the original design has definitely been outgrown by our model requirements. I have not tested it yet, will do that and post of comment of the results later, but wanted to get you the review asap.  let me know if you want to discuss my comments.  thanks!

src/ModelExecution/dataGatherer.py
```diff
@@ -136,6 +135,35 @@ def __request_dependent_data(self, dependentSeriesList: list[DependentSeries], r
...
+        """
+        # Direct hit — key is in vectorOrder
+        if key in key_to_index:
+            return key_to_index[key]
+
+        # Key is not in vectorOrder — search post-process calls for one that
+        # consumes this key as an inKey, then return the indexes of its outKey
+        for call in postProcessCalls:
+            in_keys  = [v for k, v in call.args.items() if k.endswith('_inKey')]
+            out_keys = [v for k, v in call.args.items() if k.endswith('_outKey')]
```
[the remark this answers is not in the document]
ccftissot: i agree with CP here but let's also fix the dspecs per favor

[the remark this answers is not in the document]
lovelysandlonelys: will do 👍🏻

src/ModelExecution/dataGatherer.py
lovelysandlonelys: ah, I didn't think of that at all but you're right that we don't currently do this. I'll make a note in the comment.

src/ModelExecution/dataGatherer.py
lovelysandlonelys: I added a log so that if the series isn't in the vector order or the post process list we show a warning but otherwise still go through to validate series since I think that at that point it would be a dspec issue and a warning is sufficient. if the series also fails in validation because of the nature of the unused request I think that's fine too since it'll bring attention to the series and i don't want to change the data flow too much but let me know if you think we should so something else!

data/dspec/Inundation/April/ar_inundation_apr_12h.json
```diff
@@ -68,7 +68,7 @@
             "args": {
                 "target_inKey": "DPD_12",
                 "warning_trigger": "20",
-                "outkey": "fmm_DPD_1" 
+                "outKey": "fmm_DPD_1"
```
abeasley1722, line 71: Works as intended!

abeasley1722 (approved):

bestrada33 (approved):

ccftissot:
> I think that this is actually a nice temporary solution to the problem. it does add to the argument that our ingestion process needs to be refactored; the original design has definitely been outgrown by our model requirements. I have not tested it yet, will do that and post of comment of the results later, but wanted to get you the review asap. let me know if you want to discuss my comments. thanks!

@lovelysandlonelys 
also, just occurred to me so putting it here for posterity (and in case we ever get to refactor this part), should we even date validate series that are not in output vector? might warrant a discussion.

PR #978, opened 2026-07-03
973 unit tests

### LINKED ISSUE #973: Unit Tests
Create a unit test that mocks data from each ingestion class or maybe even for each model family? To check that things will work as expected and catch edge cases.

abeasley1722:
# Goal
Create a unit test that tests each model group can produce a prediction if fed the right amount of data.

# Testing
- 'docker compose up -d --build'
- 'docker exec semaphore-core python -m pytest -s  src/tests/UnitTests/test_Semaphore.py     '

### DIFF
--- a/src/tests/UnitTests/test_Semaphore.py
+++ b/src/tests/UnitTests/test_Semaphore.py
@@ -1,0 +1,188 @@
+# -*- coding: utf-8 -*-
+# test_Semaphore.py
+# -------------------------------
+# Created By : Anointiyae Beasley
+# Created Date: 2026-06-26
+# version 1.0
+# -------------------------------
+"""
+Unit tests to test a successful run of each model group.
+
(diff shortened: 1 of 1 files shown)


lovelysandlonelys (approved):
tests run and overall seems like good code to me, testing each model group with fake data! my one comment (although not a blocker) is that the amount of print statements is... a lot and might make it difficult to identify issues if we do have a failure I tried out a way to suppress a bit of the logs but some of the statements are in other files so it doesn't work all the way :/ up to you if you have time to tinker with it but again I don't think this should be a blocker. 

```
orchestrator = Orchestrator()
    with patch.object( orchestrator, "_Orchestrator__handle_successful_prediction") as mock_success, \
         patch.object(orchestrator, "_log_input_vectors"):
```

